PR #321, opened 2023-04-02
feat: 🎸 basic impl the lsp scan

IWANABETHATGUY:
Part of https://github.com/ast-grep/ast-grep/issues/283

related vscode plugin pr is here https://github.com/ast-grep/ast-grep-vscode/pull/14


HerringtonDarkholme (approved):
LGTM!

PR #398, opened 2023-05-19
chore: use workspace package info to reduce maintenance

suyanhanx:
Use workspace package info to reduce maintenance


suyanhanx:
Seems you have an `xtask` to handle releasing tasks. 🤣

HerringtonDarkholme (approved):
Thanks! It is still a better practice to avoid duplication though!

PR #407, opened 2023-05-20
test: improve error tests

pd4d10:
Prerequisites of #392 , to make sure the styles are correct.

Before https://github.com/ast-grep/ast-grep/issues/156#issuecomment-1551768759 gets confirmed, use hard code instead.

### DIFF
--- a/crates/cli/src/error.rs
+++ b/crates/cli/src/error.rs
@@ -262,13 +262,7 @@ mod test {
       inner: &error,
     };
     let display = format!("{error_fmt}");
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme (changes requested):
Thank you for the contribution!
However I would like to change the test case a little more.

First, the test case itself is not readable. A more reasonable choice might be `RED = ansi` and wrap the expected error message in `{RED}color {_RED}`

Second, the test is for UI. We will probably change it in future so test might be frequently broken.

I would rather test if a colorized string appears in the error message.

PR #427, opened 2023-05-28
feat: supports mts and cts extensions

pd4d10:
Ref: https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions


HerringtonDarkholme (approved):
Thanks

HerringtonDarkholme (approved):
Thanks

PR #481, opened 2023-06-29
feat: Use existing language FromStr impl when Deserializing

hobofan:
- Uses the `FromStr` implementation for deserializing, which allows the aliases to be used when writing a rule YAML file
- Adds the long-form names for some of the languages to the `FromStr` impl
- Normalizes the name to lowercase so that casing doesn't matter (e.g. `tsx` vs. `Tsx` vs. `TSX`)

See also https://github.com/ast-grep/ast-grep.github.io/pull/163

### DIFF
--- a/crates/language/src/lib.rs
+++ b/crates/language/src/lib.rs
@@ -17,7 +17,7 @@ mod scala;
 use ast_grep_core::language::TSLanguage;
 use ast_grep_core::meta_var::MetaVariable;
 use ignore::types::{Types, TypesBuilder};
-use serde::{Deserialize, Serialize};
+use serde::{de, Deserialize, Deserializer, Serialize};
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme:
Hi @hobofan. Thank you for your amazing contribution! This looks really nice and solve users' confusion (sorry for not noticing it).
I'm really grateful you both helped the documentation and the CLI itself. 

Appreciate it!

hobofan:
Glad to be contributing!

I've been a fan of the concept of codemods for a long time, and I'm happy that such a great tool exists, as it helps a lot with migrations! :)

HerringtonDarkholme:
Released in 0.7.1! 
The command line also displays a list of supported languages now.

image

Issue #635, opened 2023-09-12
[feature] include rule's `note` field for `--json` output


huygn:
# ⭐ Suggestion

To include a rule's `note` field for `--json` output.

PR #636, opened 2023-09-12
feat: adds `note` field for json output

huygn:
fix https://github.com/ast-grep/ast-grep/issues/635

### DIFF
--- a/crates/cli/src/print/json_print.rs
+++ b/crates/cli/src/print/json_print.rs
@@ -449,6 +449,7 @@ id: test
 message: test rule
 severity: info
 language: TypeScript
+note: a long random note
 rule:
   pattern: "{rule}""#
       ),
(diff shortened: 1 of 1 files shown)


crates/cli/src/print/json_print.rs
```diff
@@ -474,8 +477,9 @@ rule:
       printer.print_rule(matches, file, &rule).unwrap();
       printer.after_print().unwrap();
       let json_str = get_text(&printer);
-      let json: Vec<MatchJSON> = serde_json::from_str(&json_str).unwrap();
-      assert_eq!(json[0].text, pattern, "{note}");
+      let json: Vec<RuleMatchJSON> = serde_json::from_str(&json_str).unwrap();
+      assert_eq!(json[0].matched.text, pattern, "{note}");
+      assert_eq!(json[0].note, rule.note);
```
huygn, line 482: I'm not sure if it is a good idea to update this test or to create a new test, please let me know!

HerringtonDarkholme:
Thanks! This code change looks good to me.

However, would you like to give me an example of the `note` usage? I wonder what's your use case.

huygn:
> Thanks! This code change looks good to me.
> 
> However, would you like to give me an example of the `note` usage? I wonder what's your use case.

@HerringtonDarkholme Thanks. I'm integrating `ast-grep scan` results with an internal Github App, making our PRs show scan errors as Github Annotation. currently only ruleId & message:

image

As you can see, it would be nice if I can include configured `note` field to provide guidance on how to proceed per error.

HerringtonDarkholme:
The screenshot looks fantastic! Thanks for using ast-grep in your project!

HerringtonDarkholme:
Released in 0.12.1!

PR #638, opened 2023-09-14
Add identifier convention conversions

akx:
These conversions can be tedious to write out as regexps (not to mention they were fun to implement).

```yaml
convert:
  identifier:
    from: snakecase
    to: camelcase
```
converts e.g. `snakes_live_in_forests` to `snakesLiveInTheForests`.
```yaml
convert:
  identifier:
    from: camelcase
    to: snakecase
    letterCase: capitalize
```
converts e.g. `snakesLiveInTheForests` to `Snakes_Live_In_The_Forests` (which is pretty outré, but then again hungarian notation is a thing too...).

### DIFF
--- a/crates/config/src/transform.rs
+++ b/crates/config/src/transform.rs
@@ -5,7 +5,7 @@ use ast_grep_core::{Doc, Language};
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 
-use crate::string_conversion::{apply_case_conversion, CaseConversion};
+use crate::string_conversion::{CaseConversion, IdentifierConventionConversion};
 use std::collections::HashMap;
 
 fn get_text_from_env<D: Doc>(src: &str, ctx: &mut Ctx<D>) -> Option<String> {
@@ -79,12 +79,23 @@ impl Replace {
 #[serde(rename_all = "camelCase")]
 pub struct Convert {
   source: String,
-  letter_case: CaseConversion,
+  letter_case: Option<CaseConversion>,
+  identifier: Option<IdentifierConventionConversion>,
 }
 impl Convert {
   fn compute<D: Doc>(&self, ctx: &mut Ctx<D>) -> Option<String> {
     let text = get_text_from_env(&self.source, ctx)?;
-    Some(self.letter_case.apply(text))
+    let text = if let Some(c) = self.letter_case {
+      c.apply(&text)
+    } else {
+      text
+    };
+    let text = if let Some(c) = self.identifier {
+      c.apply(&text)
+    } else {
+      text
+    };
+    Some(text)
   }
 }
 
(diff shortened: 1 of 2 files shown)


HerringtonDarkholme:
Thanks! I have also thought about this since the `convert` option is implemented! 

With regard to the API, I saw quite a lot of existing software exposing conversion without `from`. 

https://www.npmjs.com/package/stringcase
image

https://github.com/whatisinternet/Inflector

image

https://pypi.org/project/stringcase/

image

Do we need two separate `from` and `to`? 
If only `to` is needed, we can simplify the API to `convert: { letterCase: camelCase } `

crates/config/src/transform.rs
```diff
@@ -345,4 +356,22 @@ mod test {
...
   }
+
+  #[test]
+  fn test_identifier_convention_convert() -> R {
+    // Other transformations are tested in string_conversion.rs
+    let trans = parse(
+      r#"
+      convert:
+        source: "$A"
+        identifier:
```
HerringtonDarkholme, line 367: One potentially confusing part is that `identifier` will not only convert identifiers but the whole matching metavariable text. Can we make it better?

akx: My idea was that you'd capture only the identifier you care about and convert it. How do you propose this could be improved? 🤔

akx:
I was also thinking about a from-less API, but e.g. the linked `stringcase.camelcase()` e.g. just expects the input to be dash-or-underscore-separated (which isn't true for `camelCase` or `PascalCase` inputs): https://github.com/okunishinishi/python-stringcase/blob/57aae96649f5cf7fc98e50b1d62479456d492715/stringcase.py#L8-L30

I think the Zen of Python's _Explicit is Better than Implicit_ clause works here – also, now that I think of it, I'll also add a `Words` form, because doing "this is a sentence" => "thisIsASentence"  sounds useful.

HerringtonDarkholme:
I think the `identifier` field and `from` field can be both removed if 
1. We can apply a smarter separator detection, e.g. https://github.com/sindresorhus/camelcase/blob/main/index.js#L5
2. Make a more robust replacement implementation so that even the input is a sentence the output is still something readable.
Indeed the second part does not guarantee the output will be always meaningful. It still needs user to take care of applying the conversion to the appropriate ast node.

akx:
I'm okay with making `from` optional (in which case we'd just split with any separatoresque character), but I would love the option to allow the user to be explicit about it. Is that okay?

HerringtonDarkholme:
@akx 

> making from optional (in which case we'd just split with any separatoresque character), but I would love the option to allow the user to be explicit about it

This sounds great.  How about API like this?

```yaml
transform:
  B:
    convert:
      source: "$A"
      letterCase:  snakecase
      fromCase: camelcase # this is optional
```

HerringtonDarkholme:
I will first merge this request and do the remaining updates. Thanks!

HerringtonDarkholme:
I have released ast-grep 0.12.4 with these changes! API updates are recorded in https://github.com/ast-grep/ast-grep/issues/436 

Thanks!

PR #1162, opened 2024-05-23
feat: print cause of error in utils::read_file

EQt:
# Example of failure

## Before
```bash
$ ast-grep scan ./Status_IO.cpp
Cannot read file ./Status_IO.cpp
```

## After
```bash
$ ast-grep scan ./Status_IO.cpp
Cannot read file ./Status_IO.cpp: stream did not contain valid UTF-8
```

### DIFF
--- a/crates/cli/src/utils.rs
+++ b/crates/cli/src/utils.rs
@@ -187,7 +187,7 @@ fn run_worker<W: PathWorker + ?Sized + 'static>(worker: Arc<W>) -> Result<()> {
 fn read_file(path: &Path) -> Option<String> {
   let file_content = read_to_string(path)
     .with_context(|| format!("Cannot read file {}", path.to_string_lossy()))
-    .map_err(|err| eprintln!("{err}"))
+    .map_err(|err| eprintln!("{err:#}"))
     .ok()?;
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme:
Thanks for the PR! I found in the doc https://doc.rust-lang.org/std/fmt/index.html that the alternate format is `#?`.

image

Is it correct to use only `#` here?

EQt:
> Is it correct to use only `#` here?

I only read [`anyhow`'s documentation](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations) and found that you should use the alternate representation to include the cause.

Using `:?` would print the cause on multiple lines.

I'm not sure what happens if you combine both, i.e. `:#?`. Will test it.

EQt:
After looking into [`anyhow`'s source code](https://github.com/dtolnay/anyhow/blob/e6c730e8ec334eb3d7c6513940fd2dd6516ef7b6/src/fmt.rs#L10) I finally come to the conclusion that the suggestion here is the most suitable one, at least if you process a lot of source files.

HerringtonDarkholme:
thanks!

PR #1274, opened 2024-07-02
perf: use WalkBuilder::add to avoid loop

brenner8023:
## Summary by CodeRabbit

- **Refactor**
  - Improved directory traversal efficiency in the `find_util_rules` function, leading to better performance and control flow during path addition.

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -92,28 +92,30 @@ fn find_util_rules(
     return Ok(GlobalRules::default());
   };
   let mut utils = vec![];
-  // TODO: use WalkBuilder::add to avoid loop
+  let mut walker = WalkBuilder::new(&base_dir);
+
   for dir in util_dirs {
-    let dir_path = base_dir.join(dir);
-    let walker = WalkBuilder::new(&dir_path)
-      .types(config_file_type())
-      .build();
-    for dir in walker {
-      let config_file = dir.with_context(|| EC::WalkRuleDir(dir_path.clone()))?;
-      // file_type is None only if it is stdin, safe to unwrap here
-      if !config_file
-        .file_type()
-        .expect("file type should be available for non-stdin")
-        .is_file()
-      {
-        continue;
-      }
-      let path = config_file.path();
-      let file = read_to_string(path)?;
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme (approved):
Looks good

crates/cli/src/config.rs
```diff
@@ -92,28 +92,30 @@ fn find_util_rules(
     return Ok(GlobalRules::default());
   };
   let mut utils = vec![];
-  // TODO: use WalkBuilder::add to avoid loop
+  let mut walker = WalkBuilder::new(&base_dir);
```
HerringtonDarkholme, line 95: I don't think walker should iterate through base_dir by default

HerringtonDarkholme:
I will merge it first and change the WalkBuilder

HerringtonDarkholme:
image


Compile error

brenner8023:
> image
> Compile error

Fixed it. My mistake.

HerringtonDarkholme:
@brenner8023 can you also run `cargo fmt` for the commit and push again? Otherwise it will break the linting

crates/cli/src/config.rs
HerringtonDarkholme: Also by default WalkBuilder::new(base_path) will add base_path to the iteration procedure. I think we should exclude it.

brenner8023:
> @brenner8023 can you also run `cargo fmt` for the commit and push again? Otherwise it will break the linting

formatted

crates/cli/src/config.rs
```diff
@@ -92,28 +92,33 @@ fn find_util_rules(
...
-        .expect("file type should be available for non-stdin")
-        .is_file()
-      {
-        continue;
-      }
-      let path = config_file.path();
-      let file = read_to_string(path)?;
-      let new_configs = from_str(&file)?;
-      utils.push(new_configs);
+  let mut walker = WalkBuilder::new(base_dir.join(util_dirs.get(0).unwrap()));
```
HerringtonDarkholme, line 95: util_dirs can be empty so using unwarp here isn't correct

HerringtonDarkholme:
Thanks!

PR #1661, opened 2024-12-14
[RFC] feat(napi): Typed SgNode and SgRoot

mohebifar:
When using `napi`, I generally don't know what fields and kinds I'm dealing with for each language. I need to use either ast-grep or treesitter playgrounds with some sample code to know what's possible and it's generally not a good DX and might lead to (1) not covering edge cases (2) type errors

I added a script that downloads compiled treesitter node-types from the corresponding repos that's a processed version of the TS grammars that allows us to have some context about the possible kinds and fields.

I created a napi script that post-processes the index.d.ts file. Alternatively, I could introduce some proc-macros in the rust code, but that was way harder to implement and a huge headache and wasn't sure even if it's gonna work. The limitation is basically in napi not support type parameters like having it output `class SgNode<T, M>` instead of `class SgNode`. And also dynamically craft the type parameters with the macro. Happy to hop on a call on Discord to explain in more details.

https://github.com/user-attachments/assets/c7705a91-92c2-4511-be78-b0bf75a4825f

## Summary by CodeRabbit

- **New Features**
	- Enhanced type definitions for various programming languages in the API.
	- Introduced new methods for improved type safety in node handling.
	- Added a new script for generating TypeScript node type definitions.
	- New TypeScript type definitions for node schemas to enhance expressiveness.
	- Introduced constants for programming languages and their metadata.

- **Bug Fixes**
	- Updated method signatures for better clarity and type safety.

- **Documentation**
	- Modifications to the `.gitignore` to better manage TypeScript files.

- **Chores**
	- Updated package dependencies and build scripts in `package.json`.
	- Updated TypeScript configuration for faster transpilation.

### DIFF
--- a/crates/napi/package.json
+++ b/crates/napi/package.json
@@ -43,19 +43,19 @@
   "scripts": {
     "artifacts": "napi artifacts",
     "build": "napi build --no-const-enum --platform --release && yarn run typegen",
-    "build:debug": "napi build --no-const-enum --platform",
+    "build:debug": "napi build --no-const-enum --platform && yarn run typegen",
     "prepublishOnly": "napi prepublish -t npm --skip-gh-release",
     "test": "ava",
     "version": "napi version",
-    "typegen": "tsx scripts/generate-types.ts"
+    "typegen": "ts-node scripts/generate-types.ts"
   },
   "devDependencies": {
     "@napi-rs/cli": "2.18.4",
     "@types/node": "^22.10.2",
     "ava": "6.2.0",
     "chalk": "5.3.0",
+    "smol-toml": "^1.3.1",
     "ts-node": "10.9.2",
-    "tsx": "^4.19.2",
     "typescript": "5.7.2"
   },
   "ava": {
--- a/crates/napi/package.json
+++ b/crates/napi/package.json
@@ -16,7 +16,8 @@
   "files": [
     "manual.d.ts",
     "index.d.ts",
-    "index.js"
+    "index.js",
+    "types/*.ts"
   ],
   "napi": {
     "name": "ast-grep-napi",
@@ -41,17 +42,20 @@
   },
   "scripts": {
     "artifacts": "napi artifacts",
-    "build": "napi build --no-const-enum --platform --release",
+    "build": "napi build --no-const-enum --platform --release && yarn run typegen",
     "build:debug": "napi build --no-const-enum --platform",
     "prepublishOnly": "napi prepublish -t npm --skip-gh-release",
     "test": "ava",
-    "version": "napi version"
+    "version": "napi version",
+    "typegen": "tsx scripts/generate-types.ts"
   },
   "devDependencies": {
     "@napi-rs/cli": "2.18.4",
+    "@types/node": "^22.10.2",
     "ava": "6.2.0",
     "chalk": "5.3.0",
     "ts-node": "10.9.2",
+    "tsx": "^4.19.2",
     "typescript": "5.7.2"
   },
   "ava": {
--- a/crates/napi/scripts/generate-types.ts
+++ b/crates/napi/scripts/generate-types.ts
@@ -1,0 +1,264 @@
+import { readFile, writeFile } from "node:fs/promises";
+import path from "node:path";
+import { Edit, kind, Lang, parseAsync } from "../index";
+import { Rule } from "../manual";
+import { NodeTypeSchema } from "../types/node-types";
+
+const languagesNodeTypesUrls = {
+  [Lang.JavaScript]:
+    "https://raw.githubusercontent.com/tree-sitter/tree-sitter-javascript/refs/heads/master/src/node-types.json",
+  [Lang.TypeScript]:
+    "https://raw.githubusercontent.com/tree-sitter/tree-sitter-typescript/refs/heads/master/typescript/src/node-types.json",
+  [Lang.Tsx]:
+    "https://raw.githubusercontent.com/tree-sitter/tree-sitter-typescript/refs/heads/master/tsx/src/node-types.json",
+  [Lang.Java]:
+    "https://raw.githubusercontent.com/tree-sitter/tree-sitter-java/refs/heads/master/src/node-types.json",
+  [Lang.Python]:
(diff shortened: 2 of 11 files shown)


HerringtonDarkholme:
Thanks for this amazing PR! This can resolve the long-standing issue  https://github.com/ast-grep/ast-grep/issues/48

crates/napi/package.json
```diff
@@ -41,17 +42,20 @@
   },
   "scripts": {
     "artifacts": "napi artifacts",
-    "build": "napi build --no-const-enum --platform --release",
+    "build": "napi build --no-const-enum --platform --release && yarn run typegen",
     "build:debug": "napi build --no-const-enum --platform",
```
HerringtonDarkholme, line 46: We also need to add typegen here

crates/napi/package.json
```diff
@@ -41,17 +42,20 @@
...
     "prepublishOnly": "napi prepublish -t npm --skip-gh-release",
     "test": "ava",
-    "version": "napi version"
+    "version": "napi version",
+    "typegen": "tsx scripts/generate-types.ts"
```
HerringtonDarkholme, line 50: can we use ts-node? it can save several deps

crates/napi/scripts/generate-types.ts
```diff
@@ -0,0 +1,264 @@
...
+  [Lang.Yaml]:
+    "https://raw.githubusercontent.com/ikatyang/tree-sitter-yaml/refs/heads/master/src/node-types.json",
+  // Not available for SQL
+};
+
+const dirname = new URL(".", import.meta.url).pathname;
+
+async function generateLangNodeTypes() {
+  for (const [lang, url] of Object.entries(languagesNodeTypesUrls)) {
+    const nodeTypesResponse = await fetch(url);
```
HerringtonDarkholme, line 61: I agree with coderabbitai's suggestion that we need error handling here

crates/napi/scripts/generate-types.ts
```diff
@@ -0,0 +1,264 @@
+import { readFile, writeFile } from "node:fs/promises";
+import path from "node:path";
+import { Edit, kind, Lang, parseAsync } from "../index";
+import { Rule } from "../manual";
+import { NodeTypeSchema } from "../types/node-types";
+
+const languagesNodeTypesUrls = {
+  [Lang.JavaScript]:
+    "https://raw.githubusercontent.com/tree-sitter/tree-sitter-javascript/refs/heads/master/src/node-types.json",
```
HerringtonDarkholme, line 9: Is it possible to use git tag's node-types version instead of the head? ast-grep is using specific version on crates so it is not always safe to assume the git head is matching types.

crates/napi/package.json
mohebifar: Oh, sorry, I didn't notice we already have `ts-node`

crates/napi/src/sg_node.rs
```diff
@@ -75,10 +75,15 @@ impl SgNode {
     self.inner.is_named_leaf()
   }
   /// Returns the string name of the node kind
-  #[napi]
+  #[napi(ts_return_type = "T")]
   pub fn kind(&self) -> String {
     self.inner.kind().to_string()
   }
+  /// Check if the node is the same kind as the given `kind` string
+  #[napi]
```
HerringtonDarkholme, line 83: I believe napi provides macro to generate type signature so the generating script can be simplified

crates/napi/types/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.ts
```
HerringtonDarkholme, line 1: GH Copilot told me that 

> This means that all TypeScript files will be ignored except for node-types.ts.

Is this a desired change? I think the goal here is to include ts file except node-types?

crates/napi/src/sg_node.rs
mohebifar: The problem is I couldn't find any macros that allows me to specify type parameters.

I tried to use this feature as much as I could, but in certain cases like the is method, it was not possible.

What I need is `is<K extends T>(kind: K): this is SgNode<K, M>`

It's possible to specify type for the arg `kind: K`, and the return type `SgNode<..>`, but couldn't find a way to do the type parameter `<K extends T>` with napi macros.

crates/napi/types/.gitignore
mohebifar: My goal was to not to check the language-specific node types TS files in the repo as it's "auto-generated/downloaded". I don't have a strong opinion whether we should do that or not.

mohebifar: Screenshot 2024-12-14 at 11 44 34 AM

mohebifar: and `node-types` without language name prefix is the type utils to extract possible field names, etc.

crates/napi/scripts/rules.ts
```diff
@@ -0,0 +1,66 @@
...
+  kind: "class_declaration",
+  inside: {
+    kind: "ambient_declaration",
+    inside: {
+      kind: "export_statement",
+    },
+  },
+  has: {
+    field: "name",
+    regex: `^${className}$`,
```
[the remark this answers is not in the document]
mohebifar: These are not really needed as these args won't contain any special characters that would require escaping. I'd rather not complicate the script any further.

crates/napi/types/.gitignore
HerringtonDarkholme: Oh sorry I misread the .gitignore directory. It's under types and I thought it was under `napi`

crates/napi/src/sg_node.rs
HerringtonDarkholme: It's non-blocking and we can figure it out later.

HerringtonDarkholme (approved):
LGTM, let's improve this in following PRs

crates/napi/src/sg_node.rs
HerringtonDarkholme: napi-rs 3.0 supports function method type overriding but it is not available at the moment. https://github.com/napi-rs/napi-rs/pull/2362

PR #1730, opened 2025-01-10
feat: Support passing rule config to SgNode match methods (`matches`, `has`, `inside`, `follows`, `precedes`)

mohebifar:
The PyO3 API supports passing Rule as the argument to `matches`, `has`, `inside`, `follows`, `precedes` methods, but the NAPI API doesn't. I have a use case where I have the SgNode instances, and I need to check if it matches a RuleConfig and checking with just the pattern string isn't sufficient for this case.

This also makes the PyO3 and the NAPI APIs more consistent in terms of functionality.

## Summary by CodeRabbit

- **Tests**
	- Added comprehensive test cases for node matching and relationship functionalities.
	- Enhanced test coverage for parsing and node relationship validation.

- **New Features**
	- Expanded method signatures to support more flexible node matching with multiple argument types.
	- Introduced new assertions in tests to validate node structures against defined rules.

- **Improvements**
	- Increased flexibility in node matching and relationship detection.
	- Improved error handling in node-related methods.

### DIFF
--- a/crates/napi/__test__/index.spec.ts
+++ b/crates/napi/__test__/index.spec.ts
@@ -435,3 +435,52 @@ test('find node by range', t => {
   const byKind = sg.root().find(js.kind('template_string'))
   t.is(node!.id(), byKind!.id())
 })
+
+test('check if a node matches a rule using pattern', t => {
+  const sg = parse('console.log(123)')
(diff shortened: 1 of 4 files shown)


HerringtonDarkholme (approved):
Amazing!

PR #1875, opened 2025-03-13
feat(KindMatcher): Support super types

mohebifar:
As mentioned in https://github.com/ast-grep/ast-grep/pull/1784, `ast-grep` currently accepts supertypes like "declaration" and "expression" as valid types but doesn't actually match anything.  

Ideally, it should either return an `InvalidKindName` error or match all subtypes instead of doing nothing.  

Tree-sitter recently introduced support for exposing supertype data. While the tree-sitter crate already includes the necessary API changes, the parser dependencies haven't been updated in the last four months. As a result, this PR won't have any effect until the tree-sitter team updates those crates using the new `tree-sitter-cli`. We can disable the test and merge, or just wait until the `tree-sitter-typescript` crate is updated.

To make the test pass, you can replace the tree-sitter-typescript dependency in `core/Cargo.toml` with the following:

```
tree-sitter-typescript = { package = "tree-sitter-typescript-codemod", version = "0.25.0-codemod.1" }
```

## Summary by CodeRabbit

- **Chores**
  - Upgraded the `tree-sitter` library dependency to version `0.25.3`, enhancing underlying performance and stability.
- **Refactor**
  - Revamped matching functionality to support multiple subtypes more effectively, with added test coverage for supertype matching.
  - Improved language compatibility checks by switching to ABI version retrieval for better validation.
  - Standardized UTF‑16 parsing to consistently use little‑endian encoding for reliable text processing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,9 +386,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.3"
+version = "1.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "27f657647bcff5394bf56c7317665bbf790a137a50eaaa5c6bfbb9e27a518f2d"
+checksum = "be714c154be609ec7f5dad223a33bf1482fff90472de28f7362806e6d4832b8c"
 dependencies = [
  "shlex",
 ]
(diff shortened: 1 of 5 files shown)


HerringtonDarkholme:
the coverage test is failing, would you like to fix it?

mohebifar:
> the coverage test is failing, would you like to fix it?

The test is using tree-sitter-typescript and tree-sitter-typescript crate needs to be updated with the latest CLI, until then, this PR's test won't pass.

HerringtonDarkholme:
>  tree-sitter-typescript and tree-sitter-typescript crate needs to be updated with the latest CLI

sure, should we change this PR to draft now?

mohebifar:
Yeah, or we can just ignore the test and merge. Whatever you prefer. It
makes our job at Codemod easier, so we can use this feature without having
to maintain a fork and we can just swap parsers when needed (possible
because of the napi dynamic lib support).

PR #1924, opened 2025-04-12
perf: more efficient node ancestor calculation

ribru17:
When testing on the VSCode repo, with this rule:

```yaml
id: id-in-function
language: typescript
rule:
  kind: identifier
  inside:
    kind: function_declaration
    stopBy: end
```

These were the average times for `ast-grep scan --report-style=short > test.txt` with `AMD Ryzen 5 7640U`:

| Before | After |
| - | - |
| 3.9s | 2.3s |

---

Requires https://github.com/ast-grep/tree-sitter-wasm/pull/3

Explanation: Currently we iterate up the tree by getting the node's parent, in order to find ancestors. However this is slow because internally, tree-sitter does not store parent information in nodes. It retrieves parents by repeatedly calling `ts_node_child_with_descendant`, which retrieves the node's immediate child which contains the target. Instead, we can call this function directly so as to only iterate down the chain once, rather than `n` times.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the logic for hierarchical data navigation to ensure that related items are retrieved accurately and efficiently.
  - Maintained backward compatibility to preserve a seamless experience for users leveraging connected data.

### DIFF
--- a/crates/core/src/node.rs
+++ b/crates/core/src/node.rs
@@ -408,19 +408,25 @@ impl<'r, D: Doc> Node<'r, D> {
   }
 
   /// Returns all ancestors nodes of `self`.
-  /// Note: each invocation of the returned iterator is O(n)
   /// Using cursor is overkill here because adjust cursor is too expensive.
   pub fn ancestors(&self) -> impl Iterator<Item = Node<'r, D>> + '_ {
-    let mut parent = self.inner.parent();
+    let mut ancestor = Some(self.root.root().inner);
+    let self_id = self.inner.id();
     std::iter::from_fn(move || {
-      let inner = parent.clone()?;
-      let ret = Some(Node {
-        inner: inner.clone(),
+      let inner = ancestor.take()?;
+      if inner.id() == self_id {
+        return None;
+      }
+      ancestor = inner.child_with_descendant(self.inner.clone());
+      Some(Node {
+        inner,
         root: self.root,
-      });
-      parent = inner.parent();
-      ret
+      })
     })
+    // We must iterate up the tree to preserve backwards compatibility
+    .collect::<Vec<_>>()
+    .into_iter()
+    .rev()
   }
   #[must_use]
   pub fn next(&self) -> Option<Self> {


HerringtonDarkholme:
FYI tree-sitter's endiance does not support napi/pyo3 build. 

See 
https://github.com/tree-sitter/tree-sitter/discussions/4224
https://github.com/ast-grep/ast-grep/actions/runs/14431009646

ribru17:
Ah, that is unfortunate :/ does tree-sitter 0.24 support this? I believe the new method was actually added in that version

HerringtonDarkholme:
another way is to fix this one https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/portable/endian.h

to make it compiles on more platforms

ribru17:
That same build command works on my machine, maybe the docker environment does not define e.g. `__linux__`?

HerringtonDarkholme:
You can take a look at these run

https://github.com/ast-grep/ast-grep/actions/runs/14431095274/job/40465963122

https://github.com/ast-grep/ast-grep/actions/runs/14431094961

https://github.com/ast-grep/ast-grep/actions/runs/14431010713/job/40465778164


Also, I think you can try run the action manually on your fork.

image

HerringtonDarkholme:
to be more precise, tree-sitter native 0.24.7 has this method, but web-tree-sitter does not

HerringtonDarkholme:
Hi! @ribru17 I'm also working on make the system work! https://github.com/ast-grep/ast-grep/pull/1929

I am blocked by napi-rs. Hope your image can help!

ribru17:
Ah, nice! Glad to see that :D i'll see if I can help with your branch at all

.github/workflows/napi.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             target: i686-pc-windows-msvc
           - host: ubuntu-20.04
             target: x86_64-unknown-linux-gnu
-            docker: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian
+            docker: ghcr.io/ribru17/napi-rs/nodejs-rust:lts-debian
```
HerringtonDarkholme, line 46: x86_64-unknown-linux-gnu issue has been solved. aarch64-unknown-gnu is not, though

ribru17: It looks like that one is only solved because it is no longer building inside the docker container. Which could be fine, at that point maybe we just remove the docker step from aarch64 as well, unless you feel it is necessary there

HerringtonDarkholme: cross building require much more setting up so I think docker is still necessary

ribru17: I see... well I have pushed my own docker image here but CI says permission is denied. Unsure why...

HerringtonDarkholme:
it compiled
https://github.com/ast-grep/ast-grep/actions/runs/14435108453/job/40475054324

HerringtonDarkholme (approved):

HerringtonDarkholme:
inside operator does not work in wasm build

https://ast-grep.github.io/playground.html#eyJtb2RlIjoiQ29uZmlnIiwibGFuZyI6ImphdmFzY3JpcHQiLCJxdWVyeSI6ImNvbnNvbGUubG9nKCRNQVRDSCkiLCJyZXdyaXRlIjoibG9nZ2VyLmxvZygkTUFUQ0gpIiwic3RyaWN0bmVzcyI6InNtYXJ0Iiwic2VsZWN0b3IiOiIiLCJjb25maWciOiJydWxlOlxuICBraW5kOiBzdHJpbmdcbiAgaW5zaWRlOlxuICAgIGtpbmQ6IGltcG9ydF9zdGF0ZW1lbnRcbiAgICBzdG9wQnk6IGVuZCIsInNvdXJjZSI6ImltcG9ydCBhLCB7YiwgYywgZH0gZnJvbSBcIi4vZmlsZVwiO1xuaW1wb3J0IGUgZnJvbSBcIi4vb3RoZXJfZmlsZS5qc1wiO1xuaW1wb3J0IFwiLi9mb2xkZXIvXCI7XG5pbXBvcnQge3h9IGZyb20gXCJwYWNrYWdlXCI7XG5pbXBvcnQge3l9IGZyb20gXCJwYWNrYWdlL3dpdGgvcGF0aFwiO1xuXG5pbXBvcnQoXCIuL2R5bmFtaWMxXCIpO1xuaW1wb3J0KFwiLi9keW5hbWljMi5qc1wiKTtcblxubXlfZnVuYyhcIi4vdW5yZWxhdGVkX3BhdGhfc3RyaW5nXCIpXG5cbiJ9

HerringtonDarkholme:
I reverted the wasm build, but you can reproduce the issue using this snippet

https://github.com/tree-sitter/tree-sitter/tree/master/lib/binding_web

```ts
const { Parser } = require('web-tree-sitter');
await Parser.init()
const { Language } = require('web-tree-sitter');
const JavaScript = await Language.load('/path/to/tree-sitter-javascript.wasm');
parser.setLanguage(JavaScript);
const sourceCode = 'let x = 1; console.log(x);';
const tree = parser.parse(sourceCode);
const root = tree.rootNode
const child = tree.children[0].children[0]
root.childWithDescendant(child) // null
```

HerringtonDarkholme:
https://github.com/tree-sitter/tree-sitter/issues/4362

ribru17:
Ack! so frustrating, classic tree-sitter :sweat_smile: thanks for the report. I've put up a PR: https://github.com/tree-sitter/tree-sitter/pull/4363

HerringtonDarkholme:
@ribru17 thanks for the quick fix! you are amazing!

ribru17:
any time :smile: :heart:

PR #2213, opened 2025-09-03
feat(cli): impl. comp-time flag for `Commands::Docs`

MrDwarf7:
## Summary 

Very small PR that implements a feature-flag/compile time macro to strip out the `Commands::Docs` enum via `#[cfg(debug_assertions)]` as it's currently marked and handled via a `todo!()`.

Instead of having an explicit path to crash the program- preferring to hide it behind a compile time flag, this ensures that the `--release` version (aka Production) doesn't compile with unnecessary code, and doesn't have an empty path-flow that crashes the program.

## Summary by CodeRabbit

- Refactor
  - Restricted the experimental “docs” CLI subcommand to debug builds; it no longer appears or runs in production/release binaries.
  - Updated CLI help/usage so the “docs” command is omitted in release builds, reducing confusion.

- Chores
  - Streamlined build configuration to clearly separate debug-only tooling from production features, with no impact on standard release workflows.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -61,6 +61,7 @@ enum Commands {
   /// Generate shell completion script.
   Completions(CompletionsArg),
   /// Generate rule docs for current configuration. (Not Implemented Yet)
+  #[cfg(debug_assertions)]
   Docs,
 }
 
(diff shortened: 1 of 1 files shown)


MrDwarf7:
Re: Testing - I've tested this running all tests via both `cargo test --all` and `cargo test --all --workspace` etc. and we're cleanly compiling and running.

HerringtonDarkholme (approved):

PR #2274, opened 2025-10-07
perf: using ahash instead of std hash

fatelei:
fix issue #449 

## Summary by CodeRabbit

* **Performance**
  * Faster map/set operations across CLI, LSP, and libraries, improving scanning and responsiveness.
* **Refactor**
  * Swapped the underlying map/set implementation project-wide without changing behavior.
* **Chores**
  * Added a workspace dependency to enable the new map/set implementation across the workspace.
* **API**
  * A few public types now use the new map type internally; interfaces remain compatible.


HerringtonDarkholme:
@fatelei do you have any data supporting ahash is better than std hash?

fatelei:
> do you have any data supporting ahash is better than std hash?

i have no data to compare, but there is some info about this:
Large-scale string keys
On 10 M random 100-byte strings ahash builds a HashMap in ≈ 13 ms while the default hasher needs ≈ 441 ms – a ~34× speed-up .
Short-string workload (tokenised Wikipedia)
std::hash : 0.43 GiB/s
aHash     : 1.23 GiB/s (2.9× faster) 
Long-string workload (lines)
std::hash : 3.7 GiB/s
aHash     : 8.6 GiB/s (2.3× faster)

HerringtonDarkholme:
@fatelei I don't think microbenchmark will reflect real world performance for this tool. 
I would like to see performance testing on actual repositories to see if this change is worthwhile.

That's why the AHash change has not been done now. If the performance gain is negligible I won't bother introduce a new library in the tool.

PR #2310, opened 2025-10-27
fix(schemas): recognize all sgconfig fields

Dsaquel:
When creating the root `sconfig.yaml` some fields are not recognized by the yaml lsp.

This PR adds the fields that were not recognized.

## Summary by CodeRabbit

* **New Features**
  * Added custom language configuration so projects can define language variants with custom extensions and attributes.
  * Added language injection support to recognize and handle embedded languages inside host language files.
  * Added language-to-file mapping to associate languages with non-standard file extensions and syntax variants.

### DIFF
--- a/schemas/project.json
+++ b/schemas/project.json
@@ -79,6 +79,15 @@
       "properties": {
         "libraryPath": {
           "type": ["string", "object"],
+          "oneOf": [
+            { "type": "string" },
+            {
+              "type": "object",
+              "additionalProperties": {
+                "type": "string"
+              }
+            }
+          ],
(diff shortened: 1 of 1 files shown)


Dsaquel:
Thanks you for your quick review :bow: , did not see it when I pushed 1d44e8c.

HerringtonDarkholme:
Feel free to create another PR. The additionalProperties in this case is not too helpful since it is just a stringedKey value pair. But it is not harmful to add.

Dsaquel:
You right, i will abstain to do it.

PR #2317, opened 2025-10-30
fix(interactive_print): do not use the alternate screen without stdin

mathstuf:
When using `ast-grep` in scripts, the alternate screen ends up injecting ANSI sequences which can confuse tooling not expecting to emulate a TTY.

## Summary by CodeRabbit

* **New Features**
  * Added in-prompt actions for interactive highlights: quick quit or open the content in your editor after viewing.

* **Bug Fixes**
  * Improved interactive editing flow when input is piped or redirected, with alternate-screen rendering for a cleaner prompt experience and correct post-prompt behavior.

### DIFF
--- a/crates/cli/src/print/interactive_print.rs
+++ b/crates/cli/src/print/interactive_print.rs
@@ -71,18 +71,23 @@ impl InteractivePrinter {
       first_line,
       inner,
     } = highlights;
-    utils::run_in_alternate_screen(|| {
+    if self.from_stdin {
+      utils::run_in_alternate_screen(|| {
+        self.inner.process(inner)?;
+        let resp = self.prompt_view();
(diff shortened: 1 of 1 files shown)


mathstuf:
`ast-grep scan --interactive` works for me. Attached is an `asciinema` recording.

[ast-grep-interactive.txt](https://github.com/user-attachments/files/23250125/ast-grep-interactive.txt)

HerringtonDarkholme:
I tested the change.

`ast-grep -p 'test' -i` no longer prints the interactive program. This is wrong. It should output prompts like this

image


> alternate screen 

The alternate screen only impacts interactive mode and `--update-all`. So I think the correct fix should be "skip all printing when --update-all"

PR #2340, opened 2025-11-09
fix: recognize yaml extension for sgconfig file

Dsaquel:
Avoid the error:
```
Error: No ast-grep project configuration is found.
```
When creating `sgconfig` file with `yaml` extension.

## Summary by CodeRabbit

* **New Features**
  * Configuration discovery now recognizes both .yml and .yaml variants, searches upward from the current directory, and prefers the .yml variant; explicitly provided config paths remain unchanged.

* **Tests**
  * Added test coverage verifying .yaml-format configuration files are detected and used during scans.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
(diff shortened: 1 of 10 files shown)


HerringtonDarkholme:
any specific reason not using .yml extension?

Dsaquel:
Not really, it feels just a bit disrupting to not allow the yaml extention.

Dsaquel:
Just noticed that rules inside `rule_dirs` can run with .yaml extension. Maybe it make more sense that all yaml files can accept both .yml and .yaml extensions?

HerringtonDarkholme:
At least ast-grep supports yml and yaml, better than precommit lol

PR #2383, opened 2025-12-12
fix: sg completions redirect to ast-grep

Dsaquel:
Hello, my problem is that doing
```sh
sg completions fish > sg.fish
```

was outputting always `ast-grep` as the binary name.

The code in completions.rs seems to be good but the redirect for the sg aliases is undoing the binary name detection.
I don't have the whole context, maybe the fix is not right.

## Summary by CodeRabbit

* **Refactor**
  * Replaced manual subprocess spawning with a thin delegating CLI entry that forwards execution.
  * Switched to a richer error-returning style for improved error reporting and diagnostics.
  * Simplified command execution flow for easier maintenance and clearer runtime behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/cli/src/bin/alias.rs
+++ b/crates/cli/src/bin/alias.rs
@@ -1,16 +1,5 @@
-// The alias command `sg` redirects everything to ast-grep
-// we need this to avoid "multiple build target" warning
-// See https://github.com/rust-lang/cargo/issues/5930
-fn main() -> std::io::Result<()> {
-  // redirect to ast-grep
-  use std::env::args;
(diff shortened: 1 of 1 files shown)


HerringtonDarkholme:
the change will trigger the multiple build error in the comment, and the built binary size will double

Also the `sg` command is deprecated (in favor of the full command name ast-grep), due to #56 

Thanks for the contribution any way.

Dsaquel:
Thanks for the quick feedback, it is indeed better to let the developer dealing with aliases :+1: .

PR #2444, opened 2026-01-14
feat(cli): add --max-diagnostics-shown flag to scan subcommand

kioku:
## Summary

Adds `--max-diagnostics-shown=N` flag to `sg scan` that outputs at most N diagnostics and stops scanning once the limit is reached. Useful for CI pipelines where fast failure is preferred.

Workers use atomic slot reservation to coordinate the limit across parallel threads, with early termination once all slots are claimed.

## Changes

- Add `--max-diagnostics-shown` CLI argument to `ScanArg` (conflicts with `--interactive`)
- Add `should_stop()` trait method to `PathWorker` for early termination signaling
- Use atomic `fetch_update` for race-free slot reservation across parallel workers
- Truncate matches at source to ensure exact output count
- Support in both file-based and stdin modes

## Usage

```bash
# Stop after 5 diagnostics
sg scan --max-diagnostics-shown=5

# Works with stdin
echo "console.log(1); console.log(2); console.log(3);" | \
  sg scan --stdin --inline-rules '{id: no-console, language: js, rule: {pattern: "console.log($$$)"}}' \
  --max-diagnostics-shown=2
```

## Summary by CodeRabbit

* **New Features**
  * Added a --max-diagnostics-shown flag to limit number of diagnostics displayed and stop scanning once the cap is reached.

* **Tests**
  * Added tests covering the flag for multi-file, single-file, and stdin scenarios.
  * Added tests ensuring the flag conflicts with interactive mode as expected.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/cli/tests/scan_test.rs
+++ b/crates/cli/tests/scan_test.rs
@@ -407,7 +407,6 @@ fn test_max_diagnostics_shown_single_file() -> Result<()> {
 fn test_max_diagnostics_shown_stdin() -> Result<()> {
   // Test --max-diagnostics-shown with stdin input
   let input = "Some(1); Some(2); Some(3); Some(4)";
-  // Note: stdin mode returns exit code 1 when there are matches but no errors
   let output = Command::new(cargo_bin!())
     .args([
       "scan",
@@ -419,7 +418,7 @@ fn test_max_diagnostics_shown_stdin() -> Result<()> {
     ])
     .write_stdin(input)
     .assert()
-    .failure()
+    .success()
     .get_output()
     .stdout
     .clone();


HerringtonDarkholme:
Hi, thanks for contributing.

Usually for a new feature, I would recommend opening an issue first to discuss its usage, scenario and alternative design.

For this case, I agree `--max-diagnostics` is a good option to add, since most linter support such cases.

kioku:
Thanks for the feedback. Sorry about that, I'll open an issue first in the future.

Happy to discuss the design here or in a separate issue if you prefer. Let me know if you'd like any changes to the approach.

HerringtonDarkholme:
this looks good overall. I do have several changes want to do myself. I will merge this first. Thanks

PR #2480, opened 2026-02-16
chore: update schemas

Dsaquel:
When i run `cargo xtask schema` on my feature branch it change unrelated stuff.

## Summary by CodeRabbit

* **Refactor**
  * Updated position data validation to allow optional and nullable column information across all language schemas.

### DIFF
--- a/schemas/bash_rule.json
+++ b/schemas/bash_rule.json
@@ -421,14 +421,16 @@
         },
         "column": {
           "description": "0-based column number in the source code",
-          "type": "integer",
+          "type": [
+            "integer",
+            "null"
+          ],
           "format": "uint",
           "minimum": 0
         }
       },
       "required": [
-        "line",
-        "column"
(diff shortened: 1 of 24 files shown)


HerringtonDarkholme:
Thanks for the contribution. json will be formatted so this is noisy.

Dsaquel:
What do mean by that ? I did run cargo xtask schema on main not in the feature branch.

HerringtonDarkholme:
Oh, my bad.
The file will be updated during package publish. So usually you don't need manual update

Dsaquel:
Oh ok ok, but at the end it should be synced ?

Just wanted to not boat the https://github.com/ast-grep/ast-grep/pull/2482 pr of unwanted changes but lets just keep moving on 2482 i think.

HerringtonDarkholme:
I updated

PR #2482, opened 2026-02-16
feat: make default rule id to filename

Dsaquel:
Make the rule id default to the file name.

The default of `--inline-rules` will be just an empty string, i don't think we really need to address a special handling for this.

## Summary by CodeRabbit

* **New Features**
  * Rule IDs are now optional; omitted IDs are auto-generated from filenames and numbered for multi-rule files (e.g., "my-rules-0", "my-rules-1").

* **Bug Fixes**
  * Detects and reports duplicate rule IDs across configs with a clear error and appropriate exit code.

* **Tests**
  * Added comprehensive tests covering default ID generation, multi-rule numbering, mixed/explicit IDs, and duplicate-ID detection.

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -244,17 +244,15 @@ pub fn read_rule_file(
     from_yaml_string(&yaml, &Default::default())
   };
   let mut rules = parsed.with_context(|| EC::ParseRule(path.to_path_buf()))?;
-  let default_id = path
-    .file_stem()
-    .and_then(|s| s.to_str())
-    .unwrap_or("anonymous");
+  let default_id = path.file_stem().and_then(|s| s.to_str());
   let has_multiple = rules.len() > 1;
   for (i, rule) in rules.iter_mut().enumerate() {
     if rule.id.is_empty() {
+      let id = default_id.ok_or_else(|| anyhow::anyhow!(EC::InvalidRuleId(path.to_path_buf())))?;
       rule.id = if has_multiple {
-        format!("{default_id}-{i}")
+        format!("{id}-{i}")
       } else {
-        default_id.into()
+        id.into()
       };
     }
   }
--- a/crates/cli/src/utils/error_context.rs
+++ b/crates/cli/src/utils/error_context.rs
@@ -32,6 +32,7 @@ pub enum ErrorContext {
   WalkRuleDir(PathBuf),
   ReadRule(PathBuf),
   ParseRule(PathBuf),
+  InvalidRuleId(PathBuf),
   ParseTest(PathBuf),
   InvalidGlobalUtils,
   GlobPattern,
@@ -83,7 +84,7 @@ impl ErrorContext {
       ReadConfiguration | ReadRule(_) | WalkRuleDir(_) | WriteFile(_) => 6,
       StdInIsNotInteractive => 7,
       ParseTest(_) | ParseRule(_) | ParseConfiguration | ParsePattern | InvalidGlobalUtils
-      | LangInjection | DuplicateRuleId(_) => 8,
+      | LangInjection | DuplicateRuleId(_) | InvalidRuleId(_) => 8,
       GlobPattern | BuildGlobs => 9,
       CannotInferShell => 10,
       ProjectAlreadyExist | FileAlreadyExist(_) => 17,
@@ -154,6 +155,11 @@ impl ErrorMessage {
       ParseRule(file) => Self::new(
         format!("Cannot parse rule {}", file.display()),
         "The file is not a valid ast-grep rule. Please refer to doc and fix the error.",
+        CONFIG_GUIDE,
+      ),
+      InvalidRuleId(file) => Self::new(
+        format!("Cannot infer rule id from {}", file.display()),
+        "The rule file name is not valid UTF-8. Please add an explicit `id` field to the rule.",
         CONFIG_GUIDE,
       ),
       GlobPattern => Self::new(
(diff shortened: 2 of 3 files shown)


Dsaquel:
@coderabbitai review

crates/cli/src/config.rs
```diff
@@ -229,7 +229,21 @@ pub fn read_rule_file(
   } else {
     from_yaml_string(&yaml, &Default::default())
   };
-  parsed.with_context(|| EC::ParseRule(path.to_path_buf()))
+  let mut rules = parsed.with_context(|| EC::ParseRule(path.to_path_buf()))?;
+  let missing_id_count = rules.iter().filter(|c| c.id.is_empty()).count();
+  if missing_id_count > 1 {
```
Dsaquel: I didn't know but a rule file can contains multiple rule in it through this syntax:

```yml
  language: TypeScript
  rule: { pattern: Some($A) }
  message: first rule
  ---
  language: TypeScript
  rule: { pattern: None }
  message: second rule
```

In this case, id is required (`Ec::MissingRuleId`).

crates/cli/src/utils/error_context.rs
```diff
@@ -155,6 +156,11 @@ impl ErrorMessage {
         "The file is not a valid ast-grep rule. Please refer to doc and fix the error.",
         CONFIG_GUIDE,
       ),
+      MissingRuleId(file) => Self::new(
```
HerringtonDarkholme: Do we need this one? We can assign a number to each rule in multiple-rule YAML

Dsaquel: It was my first attempt actually :laughing: , ok i'll update.

crates/cli/src/config.rs
Dsaquel: Already replied below, removing this behavior.

crates/cli/src/utils/error_context.rs
Dsaquel: Updated [050063d](https://github.com/ast-grep/ast-grep/pull/2482/commits/050063d10a4a6d6f440aca0821d881fa21b0e503).

crates/cli/src/utils/error_context.rs
```diff
@@ -32,6 +32,7 @@ pub enum ErrorContext {
   WalkRuleDir(PathBuf),
   ReadRule(PathBuf),
   ParseRule(PathBuf),
+
```
HerringtonDarkholme: nit: remove

crates/cli/src/config.rs
```diff
@@ -229,7 +229,21 @@ pub fn read_rule_file(
...
+  let default_id = path
+    .file_stem()
+    .and_then(|s| s.to_str())
+    .unwrap_or_default();
+  for rule in &mut rules {
+    if rule.id.is_empty() {
+      rule.id = default_id.to_string();
+    }
+  }
+  Ok(rules)
```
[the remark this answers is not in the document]
HerringtonDarkholme: Correct, @Dsaquel can you fix this? by adding a annoymous something?

HerringtonDarkholme (changes requested):
Can you also add `--inline-rules` test? thanks

Dsaquel:
Thanks for your review i'll update.

crates/cli/src/utils/error_context.rs
Dsaquel: addressed c196184c7ea9

Dsaquel:
>Can you also add --inline-rules test? thanks

Updated through 14274dcb4b9b.

crates/cli/src/config.rs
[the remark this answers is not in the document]
Dsaquel: updated through a421cadc5fa4.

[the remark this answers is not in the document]
Dsaquel: @coderabbitai review again pls

crates/cli/src/config.rs
```diff
@@ -229,7 +229,22 @@ pub fn read_rule_file(
   } else {
     from_yaml_string(&yaml, &Default::default())
   };
-  parsed.with_context(|| EC::ParseRule(path.to_path_buf()))
+  let mut rules = parsed.with_context(|| EC::ParseRule(path.to_path_buf()))?;
+  let default_id = path
+    .file_stem()
+    .and_then(|s| s.to_str())
+    .unwrap_or("anonymous");
```
HerringtonDarkholme: Consider this, if a user has multiple YAML files without id, this code will result in multiple same name rule and trigger an error. Would you like to add this case to integration test?

Dsaquel: Yes sure, make sense.

Dsaquel: Just to make sure: having `toto.yaml` and `toto.yml` without ids will make no error do you ask me to implement this error and write a test case ?

Dsaquel: Was not sure so i implemented it. 784d5e46fd66.

HerringtonDarkholme: My point is, if ast-grep is providing default id, it should not trigger unexpected behavior.
e.g. later unnamed yaml rule overriding previous unnamed yaml rule.

In this case, if the default id is not available due to wrong filename stem, it should not implicitly override later unnamed rule.

In this fix, `anonymous` will make later YAML file overridden. This is an implicit behavior.

There are two options to fix this
* create unique anonymous id for each unnamable rule
* return an error when the id cannot be inferred

Dsaquel: Thanks for the clarification, i have addressed your request change in 381f8912bd72. We throw an error if the rule id cannot be inferred from the filename.

PR #2516, opened 2026-03-05
Add Erlang support

robertoaloi:
Add support for the [Erlang](https://www.erlang.org/) programming language, using WhatsApp's [tree-sitter-erlang](https://crates.io/crates/tree-sitter-erlang) grammar.

```
cargo test -p ast-grep-language
cargo run -- -p '{ok, $B}' -l erlang /path/to/file.erl
```

## Summary by CodeRabbit

* **New Features**
  * Added support for the Erlang language, enabling pattern matching and code transformations on .erl and .hrl files.

* **Tests**
  * Added unit tests covering Erlang matching, non-matching, and replacement scenarios.

* **Chores**
  * Extended schema generation and build/test mappings to include Erlang outputs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,7 @@ dependencies = [
  "tree-sitter-cpp",
  "tree-sitter-css",
  "tree-sitter-elixir",
+ "tree-sitter-erlang",
  "tree-sitter-go",
  "tree-sitter-haskell",
  "tree-sitter-hcl",
@@ -2117,6 +2118,16 @@ name = "tree-sitter-elixir"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "66dd064a762ed95bfc29857fa3cb7403bb1e5cb88112de0f6341b7e47284ba40"
+dependencies = [
+ "cc",
+ "tree-sitter-language",
+]
+
+[[package]]
+name = "tree-sitter-erlang"
+version = "0.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2091cce4eda19c03d77928c608ac6617445a6a25691dde1e93ac0102467a6be"
 dependencies = [
  "cc",
  "tree-sitter-language",
--- a/crates/language/Cargo.toml
+++ b/crates/language/Cargo.toml
@@ -23,6 +23,7 @@ tree-sitter-c-sharp = { version = "0.23.0", optional = true }
 tree-sitter-css = { version = "0.25.0", optional = true }
 tree-sitter-c = { version = "0.24.0", optional = true }
 tree-sitter-elixir = { version = "0.3.0", optional = true }
+tree-sitter-erlang = { version = "0.15.0", optional = true }
 tree-sitter-go = { version = "0.25.0", optional = true }
 tree-sitter-haskell = { version = "0.23.0", optional = true }
 tree-sitter-html = { version = "0.23.0", optional = true }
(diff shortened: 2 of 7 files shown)


HerringtonDarkholme:
Hi, thanks for the contribution. 

tree-sitter-erlang is maintained by what's app, who is actively using erlang. So I assume the crate quality should be okay.

That said, language popularity and grammar size should still be considered. https://ast-grep.github.io/contributing/add-lang.html#requirements-and-constraints


Is whatsapp actively using ast-grep? Alternatively, can the custom language configuration meet your use case?

robertoaloi:
Hi @HerringtonDarkholme 

I can confirm the tree-sitter grammar is high quality and actively maintained by our team at WhatsApp. The grammar is the basis for the [ELP language server](https://github.com/WhatsApp/erlang-language-platform/), the main language server for Erlang. The language server (and therefore the grammar) is [used by the same team](https://github.com/erlang/otp/blob/master/.elp.toml) that maintains the Erlang language at Ericsson. It's also used in other significant projects such as [difftastic](https://github.com/Wilfred/difftastic) - over 20K stars on GitHub.

Erlang is # 36 in the Tiobe index. Haskell, which is part of ast-grep already, is at # 35. Scala (also part of ast-grep) is at # 39. Elixir - also part of ast-grep, is at # 45. Elixir code compiles into Erlang and both share the same virtual machine, the BEAM virtual machine. Elixir dependencies are often Erlang libraries.

I think Erlang would be a fair addition to the library and would enable interesting use cases.

robertoaloi:
(closing was not intentional)

HerringtonDarkholme:
I'm asking if Erlang community is actually using this tool.

A lot of langauges mentioned above are adopted in early day of ast-grep, esp Elixir. They are added, but no one is actually using them I believe. Dart was added before and removed later, yet no one complained about the removal.

Adding new langauges is adding more maintenance burden to this project. HCL is a prominent example. It is added by external contributors and was requested by multiple people. However, it still lacks things like playground support, testing and random ppl will jump out complaining missing features or confusing parsing behaviors.

Also, Tiobe is not the only criterion. Searching GitHub showed Scala has 9.3M files, Haskell has 2M files. But Erlang has only 500K files. TypeScript, the most popular language these days, ranks only 35 on TIOBE.

I would consider adding niche languages if any significant programming language organization is endorsing, or at least being aware of this tool. Otherwise a niche tool for a niche community won't make a big impact IMHO. 

And you can readily use the tool using custom language loading. I need more evidence that having the language builtin is making more adoption than having it as custom language.

robertoaloi:
> I'm asking if Erlang community is actually using this tool.

I assume not yet, since out-of-the-box support the language was not available until this PR.

The main reason I am interested in native support for Erlang in `ast-grep` is that the ELP language server for Erlang (also written in Rust and leveraging tree-sitter) offers a huge number of linters. ELP currently has its own mechanism to match and transform ASTs, named [SSR](https://whatsapp.github.io/erlang-language-platform/docs/contributing/linters/linter-traits/#ssr-pattern-syntax) (Semantic Search and Replace). While SSR offers various advantages over `ast-grep` (as an example, it allows to deal with macro expansions), SSR Is still incomplete in many ways. And `ast-grep` could serve well to write many linters, and provide auto-fixes.

My assumption here is that the *custom language configuration* would make it harder to include `ast-grep` with Erlang support as a dependency for ELP. If this is not correct, then a native integration would not strictly be required.

One option could be to add Erlang support and experiment with the integration in ELP. If that turns out successful, the entire Erlang community will basically be using ast-grep. If not, we could drop support for it.

HerringtonDarkholme:
@robertoaloi Can you try https://ast-grep.github.io/advanced/custom-language.html to colllect feedbacks from the community? 
ast-grep already supports loading language dynamically so the core doesn't need to ship or unship a language for testing out.

LuuOW (commented):
Technical audit: code patterns and functional logic verified for system integrity.

PR #2577, opened 2026-04-02
fix(napi): prevent SIGSEGV when GC collects SgRoot during concurrent async ops

hieuit095:
## Bug Description

@ast-grep/napi crashes with SIGSEGV when Node.js garbage-collects native AST objects (SgRoot/SgNode) while concurrent async operations are in-flight. The crash occurs in the native destructor during GC, not during parsing itself.

Reproduction: parse 11,000+ PHP files, then make concurrent HTTP calls. The GC runs during the async phase and collects SgRoot objects, causing a SIGSEGV when their 	ree_sitter::Tree is freed.

## Root Cause

SgNode uses SharedReference<SgRoot, NodeMatch> which does **not** keep SgRoot alive. When SgRoot is the only reference to the JavaScript object (i.e., the JS caller drops the reference from parse()), the SharedReference still borrows from a weak Reference that was immediately dropped. This allows GC to collect SgRoot while SgNode objects still reference its 	ree_sitter::Tree.

## Fix

Store an explicit owner: Option<Reference<SgRoot>> in each SgNode. This maintains a strong Reference that prevents GC collection of SgRoot for the lifetime of all derived SgNode objects. The #[napi(skip)] attribute ensures this field is Rust-internal only and not exposed to the JavaScript API.

## Changes

- crates/napi/src/sg_node.rs: Added owner field to SgNode struct and updated all SgNode constructors to populate it
- crates/napi/src/find_files.rs: Set owner: None for ind_files code path (internal-only, not exposed to JS)

## Testing

All existing unit tests pass (cargo test -p ast-grep-napi).

## Summary by CodeRabbit

* **Refactor**
  * Improved internal node ownership and lifecycle handling to increase stability and reduce risk of resource leaks. No user-facing behavior or public APIs changed.

### DIFF
--- a/crates/napi/src/find_files.rs
+++ b/crates/napi/src/find_files.rs
@@ -1,231 +1,152 @@
-use ast_grep_config::RuleCore;
-use ast_grep_core::pinned::{NodeData, PinnedNodeData};
-use ast_grep_core::{AstGrep, NodeMatch};
-use ignore::{WalkBuilder, WalkParallel, WalkState};
-use napi::anyhow::{anyhow, Context, Result as Ret};
-use napi::bindgen_prelude::*;
-use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
-use napi::Task;
-use napi_derive::napi;
-use std::collections::HashMap;
-use std::sync::atomic::{AtomicU32, Ordering};
-
-use crate::doc::{JsDoc, NapiConfig};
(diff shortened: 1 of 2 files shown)


HerringtonDarkholme:
is this the root cause? can you provide a repro that demonstrates the sigsegv problem?

HerringtonDarkholme:
I don't think this fixes the problem, see comments in #2553

PR #2595, opened 2026-04-16
fix: replace postinstall script with binary resolution in ast-grep

mohebifar:
Modern package managers like pnpm v10 disable postinstall scripts by default. This PR just updates the cli npm package to follow Codemod CLI's child_process pattern: https://github.com/codemod/codemod/tree/main/crates/cli/npm

## Summary by CodeRabbit

* **Refactor**
  * Improved native binary discovery and execution: better platform detection, robust fallback lookup, signal forwarding, and clearer error/exit behavior.

* **Chores**
  * Removed post-install installation steps and simplified package publish contents.
  * Made the command-line wrapper an executable entrypoint that delegates to the native runner.


HerringtonDarkholme:
I still prefer to have a postinstall script to accelerate the startup time. 
Note the nodejs starting up is not negligible compared to ast-grep's runtime (say, 200ms starting node vs 100ms actual search)

Preferably, add a gentle note in JS script is also helpful to nudge users to use approve-builds

PR #2607, opened 2026-04-22
fix(lsp): report unused suppressions without rules

### LINKED ISSUE #2530: fix(lsp): unused-suppression hints silently lost for languages with no configured rules
## Problem

In `crates/lsp/src/lib.rs`, the `get_diagnostics` method iterates over the host document and its injected language regions. When `rule_refs.is_empty()` for a given language, the loop `continue`s before `CombinedScan::set_unused_suppression_rule` is called.

This means that if a host or injected language region has **no configured rules**, unused-suppression diagnostics (hints) are silently dropped for that language — even when there are suppression comments that are now unused.

## Steps to Reproduce

1. Configure rules for only one of two languages present in a file (e.g., TypeScript injected into HTML).
2. Add a suppression comment for the language with no rules.
3. Observe that no "unused suppression" hint is reported for the suppression comment.

## Expected Behavior

Unused-suppression hints should still be emitted even for languages that have zero configured rules, because `CombinedScan::unused_config` produces a standalone rule for detecting unused suppressions — independent of user-defined rules.

## Related

- PR: https://github.com/ast-grep/ast-grep/pull/2528
- Review comment: https://github.com/ast-grep/ast-grep/pull/2528#discussion_r2914067518

Reported by @HerringtonDarkholme.

lawrence3699:
fix #2530

When an injected language has no configured rules, `get_diagnostics` skips that region before it sets up the standalone unused-suppression rule.

This change scans host and injected documents even when their rule set is empty, so unused `ast-grep-ignore` hints still show up for those regions.

Validation:
- cargo fmt --all --check
- cargo test -p ast-grep-lsp
- cargo test
- cargo clippy --all-targets --all-features --workspace --release --locked -- -D clippy::all

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected detection of suppression comments in embedded/injected code so unused suppressions are reported even when the embedded language has no configured rules.
* **Tests**
  * Added an automated test to ensure unused suppression warnings are emitted for injected language blocks without rules.

### DIFF
--- a/crates/lsp/src/lib.rs
+++ b/crates/lsp/src/lib.rs
@@ -285,16 +285,9 @@ impl<L: LSPLang> Backend<L> {
     let mut fixes = Fixes::new();
     let injections = root.get_injections(|lang| L::from_str(lang).ok());
     let docs = std::iter::once(root).chain(injections.iter());
-    let doc_and_rules = docs.filter_map(|injected| {
-      let rule_refs = rules.get_rule_from_lang(&path, injected.lang().clone());
-      if rule_refs.is_empty() {
-        None
-      } else {
-        Some((injected, rule_refs))
-      }
-    });
     // iterate over all main doc and injected docs
-    for (injected, rule_refs) in doc_and_rules {
+    for injected in docs {
+      let rule_refs = rules.get_rule_from_lang(&path, injected.lang().clone());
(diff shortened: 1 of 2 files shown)


HerringtonDarkholme:
Thanks!

Issue #2638, opened 2026-05-12
Bug: LSP diagnostics lost during initialization — publishDiagnostics suppressed before server enters Initialized state


Zireael:
## Bug Description

**Component:** LSP Server — Diagnostics Delivery
**Summary:** `tower-lsp-server`'s `send_notification` silently suppresses all notifications (including `textDocument/publishDiagnostics`) when the server has not yet entered the `Initialized` state. This causes diagnostics from the first `didOpen` to be lost without any feedback.

## How It Happens

The race condition occurs when:

1. Client sends `initialize` request
2. Server responds with capabilities
3. Client sends `initialized` notification
4. Client sends `textDocument/didOpen` notification
5. Server starts analysis and calls `publish_diagnostics`
6. **BUT** if `publish_diagnostics` is called before the server's internal state transitions to `Initialized`, tower-lsp silently drops it

In `tower-lsp-server/src/service/client.rs`:

```rust
async fn send_notification<N>(&self, params: N::Params) where N: notification::Notification {
    if let State::Initialized | State::ShutDown = self.inner.state.get() {
        self.send_notification_unchecked::<N>(params).await;
    } else {
        let msg = Request::from_notification::<N>(params);
        trace!("server not initialized, supressing message: {}", msg);
        // ↑ This is invisible because there's no tracing subscriber (see #2637)
    }
}
```

## Evidence

In ast-grep's LSP backend (`crates/lsp/src/lib.rs`):

```rust
async fn did_open(&self, params: DidOpenTextDocumentParams) {
    // ... parse and analyze document ...
    // This calls publish_diagnostics which goes through send_notification
    self.publish_diagnostics(uri, version, diagnostics).await;
}
```

If `did_open` is processed before the `Initialized` state transition completes, the diagnostics are lost. Since there's no tracing subscriber (as documented in #2637), the `trace!` log is invisible.

## Reproduction

This is difficult to reproduce deterministically because it depends on the timing between the `initialized` notification processing and the `did_open` processing. It's more likely to occur when:

- The LSP server is under load (slow initialization)
- The client sends `didOpen` very quickly after `initialized`
- The server is processing multiple files at startup (workspace indexing)

## Impact

- Users see **no diagnostics** for files opened immediately after LSP startup
- The issue is intermittent and timing-dependent
- There's **zero visibility** into the dropped diagnostics (no logging, no error)
- Users may assume ast-grep LSP doesn't work and disable it

### Option A: Buffer notifications until initialized

[...]

PR #2640, opened 2026-05-12
fix(lsp): buffer diagnostics during initialization race (#2638)

Zireael:
## Summary

tower-lsp-server's `send_notification` silently drops `textDocument/publishDiagnostics` when the server has not yet reached `State::Initialized`. Since `didOpen` arrives asynchronously after `initialized`, diagnostics for the first opened file are silently lost.

This fix adds a buffer inside the `Backend` struct: diagnostics published before `initialized()` runs are queued, then flushed once the server enters the Initialized state. The `initialized` flag is set before `reload_rules()` so that any diagnostics triggered during rule loading also flow through.

## Changes

- `crates/lsp/src/lib.rs`: added `initialized: AtomicBool` and `pending_diagnostics: Mutex<Vec<(Uri, i32, Vec<Diagnostic>)>>` to `Backend`. Guarded `publish_diagnostics` to buffer when not yet initialized. Set the flag and drain the buffer at the top of `initialized()`. Replaced `eprintln!` debug lines with `tracing::debug!`.

## Test plan

`cargo test -p ast-grep-lsp` — all 11 integration tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Diagnostics are now buffered and delivered after the language server finishes initialization, preventing premature diagnostic delivery.

* **Chores**
  * Initialized logger/tracing on the language server and added debug logging to show rule-scan and match counts for better visibility.

* **Breaking Changes**
  * Public language type trait now requires Debug, which may affect downstream implementations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ast-grep/ast-grep/pull/2640)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,7 @@ dependencies = [
  "termimad",
  "terminal-light",
  "tokio",
+ "tracing-subscriber",
  "tree-sitter",
 ]
 
@@ -228,6 +229,7 @@ dependencies = [
  "tokio-stream",
  "tokio-util",
  "tower-lsp-server",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


HerringtonDarkholme (changes requested):
is this a real issue when users are using?

Otherwise I incline to not merge this since it bloats the code with additional atomic field and buffer

HerringtonDarkholme:
Per LSP spec client should NOT send any request to server, it is okay to ignore the request and I don't want to add new untested code for such edge case. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

image

Issue #2000, opened 2025-05-20
[feature] Open in playground feature


samwightt:
# ⭐ Suggestion

It would be nice if there were a CLI command and / or way in VSCode to open a rule and the error that triggered it in the playground.

For example, maybe the command is:

```sh
ast-grep open-playground --file ./something.ts --rule some-rule-id # Errors show the path and the rule ID,
# not the rule file location.

# Also I don't like this command name, it's just an example. Might be nice to put it under a 'debug' command group or something.
```

The command would use [`open`](https://docs.rs/open/latest/open/) and open the link in the browser.

# 💻 Use Cases

Sometimes when there's an error brought up by a rule I want to be able to open the playground with the rule and the code file to see what's going on. I can do this manually, but it'd be nice if there was a shortcut.

Testing Library has this as a feature, where if you add [`screen.logTestingPlaygroundURL()`](https://testing-library.com/docs/dom-testing-library/api-debugging/#screenlogtestingplaygroundurl) to a test it will output a clickable URL to the testing library playground.

PR #2650, opened 2026-05-19
feat(cli): add `ast-grep playground` subcommand

younggglcy:
## Summary

Adds a top-level `ast-grep playground` subcommand that opens a rule and/or source file in the [web playground](https://ast-grep.github.io/playground.html) with state preloaded via a base64-encoded URL fragment.

Refs #2000.

## Usage

```sh
ast-grep playground --file path/to/source.ts
ast-grep playground --rule no-console
ast-grep playground --rule-file ./my-rule.yml
ast-grep playground --file a.ts --rule no-console
ast-grep playground --file a.ts --print      # print URL only, no browser open
```

## Behaviour

- Encoding matches the playground frontend's `btoa(unescape(encodeURIComponent(JSON.stringify(state))))` byte-for-byte (see `website/src/utils.ts`), so the URL is bit-compatible with what the playground produces from its own UI.
- `--rule <ID>` walks the project's `sgconfig.yml`, looks up the rule, and inlines any reachable global utilities into the shared rule's `utils:` block so the URL is self-contained and parses standalone via `from_yaml_string`. Parameterised utilities, language mismatches, and utilities using `constraints` / `transform` / `fix` are reported as errors.
- Inlined project utilities have their metavariables renamed to `AST_GREP_PLAYGROUND_<UTIL_ID>_<NAME>` so an inlined util's `$A` does not collide with the caller's `$A`. Variadic (`$$$X`) and anonymous (`$_`, `$$$_`) forms are preserved.
- Default flow opens the URL in the user's default browser and stays quiet on success; on failure it falls back to printing the URL on stdout. `--print` skips the browser and writes only the URL to stdout (script-friendly).
- Status messages on stderr respect `NO_COLOR` / `TERM=dumb` / TTY detection, mirroring `env_allows_color` in `crates/cli/src/print/colored_print/styles.rs`.
- Refuses to share rules/files in languages not supported by the web playground (Dart / Haskell / Solidity / custom languages) rather than silently letting the frontend fall back to JavaScript.
- Empty optional fields are stripped from the serialised rule YAML via a structural `serde_yaml::Value` walk, so block scalars containing the literal text `foo: null` survive.

New dependencies: `open = "5"` (default-browser launcher) and `base64 = "0.22"`.

Exposes `ProjectConfig::find_util_rule_configs()` to share the intermediate `Vec<SerializableGlobalRule<SgLang>>` with the existing `find_util_rules`; otherwise no public-API churn.

## Test plan

- [x] `cargo test -p ast-grep --lib playground::tests` — 20 unit tests covering URL encoding (incl. unicode source), language resolution priority, source/rule resolution, null-line stripping, project-util inlining, and metavar scope preservation
- [x] `cargo test -p ast-grep --test playground_test` — 5 integration tests via `assert_cmd` covering `--file`, `--rule-file`, `--rule <id>` with a fixture `sgconfig.yml`, `--rule` vs `--rule-file` mutual exclusion, and the "nothing to share" error
- [x] `cargo test -p ast-grep` — full crate test suite passes
- [x] `cargo clippy -p ast-grep --all-targets -- -D warnings` — clean
- [x] `cargo fmt -p ast-grep` — clean
- [x] Manual smoke: `cargo run -- playground --file <some-source>` opens the browser to the playground with the source preloaded; `--rule <id>` from inside a project config loads the rule in Config mode

## Summary by CodeRabbit

* **New Features**
  * New playground subcommand to generate a shareable playground URL from rules, configs, or source files; can open the playground in your browser or print the URL.
  * Language inference for shared state (flag, rule metadata, or file extension).

* **Behavior Changes**
  * Shared rule YAML must be self-contained for sharing; project utility references are rejected.

* **Bug Fixes**
  * Clearer errors when reading/parsing rule files, for conflicting options, unsupported languages, and "nothing to share".

[Review Change Stack](https://app.coderabbit.ai/change-stack/ast-grep/ast-grep/pull/2650?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/cli/src/playground.rs
+++ b/crates/cli/src/playground.rs
@@ -1,0 +1,947 @@
+use std::fs::read_to_string;
+use std::path::{Path, PathBuf};
+use std::process::ExitCode;
+use std::str::FromStr;
+
+use ansi_term::{Color, Style};
+use anyhow::{anyhow, Result};
+use ast_grep_config::{RuleConfig, SerializableGlobalRule};
+use ast_grep_core::Language;
+use ast_grep_language::SupportLang;
+use base64::engine::general_purpose::STANDARD as B64;
+use base64::Engine;
+use clap::Args;
+use crossterm::tty::IsTty;
+use serde::{Deserialize, Serialize};
+use serde_yaml::{Mapping, Value};
+use std::collections::{HashMap, HashSet, VecDeque};
+
+use crate::config::ProjectConfig;
+use crate::lang::SgLang;
+use crate::utils::RuleOverwrite;
+
+const PLAYGROUND_BASE: &str = "https://ast-grep.github.io/playground.html";
+
+/// `ansi_term` styles applied to the `--print`-less stderr status lines.
+#[derive(Default)]
+struct StatusStyles {
+  info: Style,
+  success: Style,
+  warning: Style,
+}
+
+impl StatusStyles {
+  /// Pick colored styles when stderr is a TTY and `NO_COLOR`/`TERM=dumb`
+  /// aren't set; fall back to plain `Style::default()` otherwise.
+  fn auto() -> Self {
+    let no_color = std::env::var_os("NO_COLOR").is_some();
+    let dumb = std::env::var_os("TERM")
+      .map(|t| t == "dumb")
+      .unwrap_or(false);
+    if no_color || dumb || !std::io::stderr().is_tty() {
+      Self::default()
+    } else {
+      Self {
+        info: Style::new().dimmed(),
+        success: Style::new().fg(Color::Green),
+        warning: Style::new().fg(Color::Yellow),
+      }
+    }
+  }
+}
+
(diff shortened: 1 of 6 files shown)


crates/cli/src/playground.rs
```diff
@@ -0,0 +1,947 @@
...
+    rule_map.insert(Value::String("utils".into()), Value::Mapping(utils));
+  }
+  Ok(())
+}
+
+/// Convert a project-level global utility into a value suitable for the
+/// playground's local `utils:` block. Rejects utilities that can't be
+/// represented locally (parameterized, language-mismatched, or carrying
+/// constraints / transforms / fix that aren't supported on local utils).
+fn project_util_to_local_util(
```
HerringtonDarkholme: I won't consider including such usage in the playground command. it is way to complicated to do the inline-patching

PR #2661, opened 2026-05-23
Add TOML support

jayvdb:
## Release Notes

* **New Features**
  * Added TOML language support for pattern matching and rule definitions

* **Bug Fixes**
  * Improved literal value matching to correctly distinguish between different values (e.g., "foo" vs "bar" strings, 100 vs 200 numbers) across JSON, YAML, HTML, Kotlin, Rust, and TOML
  * Fixed YAML scalar node handling to properly preserve empty and non-empty literal text distinction

[Review Change Stack](https://app.coderabbit.ai/change-stack/ast-grep/ast-grep/pull/2661?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,7 @@ dependencies = [
  "tree-sitter-scala",
  "tree-sitter-solidity",
  "tree-sitter-swift",
+ "tree-sitter-toml-ng",
  "tree-sitter-typescript",
  "tree-sitter-yaml",
 ]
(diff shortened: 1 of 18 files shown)


jayvdb:
@coderabbitai review

jayvdb:
While adding TOML support, with lots of tests, I found some problems with YAML, and then one also with Kotlin (but I haven't checked this one carefully).  Happy to split the YAML & Kotlin fixes to a separate PR, however the way to fix those is tightly related to the TOML problems.

This has been mostly constructed using Claude, so some cleanup is still required, but it would be helpful to get some feedback before finalising this PR.  It tried a few approaches at solving these problems, so I left the various approaches in separate commits so the reviewers can see the various approaches if they want.

HerringtonDarkholme:
Please add discussion here before opening a language PR
https://github.com/ast-grep/ast-grep/discussions/419

jayvdb:
@HerringtonDarkholme that step is not mentioned on https://ast-grep.github.io/contributing/add-lang.html , and this PR is about way more than adding TOML.  See my comment above explaining this.

PR #2676, opened 2026-06-02
fix: don't leak metavar bindings from a negated `not` rule

knutwannheden:
> **Disclaimer:** This PR was created with the help of [Claude Code](https://claude.com/claude-code). The root-cause investigation, fix, and regression tests were produced with AI assistance and reviewed by me before submitting.

## Motivation

A `not` rule must never contribute metavariable bindings — a successful negation means the inner rule did **not** match, so there is nothing to bind. But `Not::match_node_with_env` passed the **live** metavariable environment straight to its inner rule. When the inner *matches* (which makes the negation fail), it writes its bindings into the env and `Not` returns `None` **without reverting them**.

This is normally masked: `all`/`any` match into a cloned env and discard it when a branch fails, so a leaked binding never escapes. But a **relational rule** (`has`/`inside`/`precedes`/`follows`) evaluates its inner via `find_map` over many candidate nodes, reusing a **single** env with no clone in between. So a binding leaked by an early candidate (one where the negated inner matched) survives into a later, **successful** match and lands in the result env — corrupting any `fix`/`transform`/`constraints` that references the variable.

This is the same env-leak class fixed for the ellipsis-end lookahead probe in #2670, in a different matcher.

## Example

Target:

```js
function f() { bar(); return foo; target; }
```

Rule:

```yaml
id: leak
language: typescript
rule:
  kind: expression_statement
  regex: '^target'
  follows:
    not:
      pattern: return $A
    stopBy: end
```

`follows` walks the preceding siblings nearest-first. The nearest one, `return foo;`, matches the negated `return $A` (binding `$A = foo`), so the negation fails there and the leaked `$A = foo` stays in the env. A farther sibling, `bar();`, is not a `return`, so the negation succeeds and the node matches.

The `not` is meant to bind nothing, yet before this change the match's env contains `$A = foo`. After it, `$A` is correctly unbound.

## Summary

- In `Op::Not::match_node_with_env`, run the negated inner against a throwaway `Cow::Borrowed` clone of the env instead of the live one, so an inner match can never leak partial bindings into the caller's environment.
- Add regression tests at three levels: the `Not` operator directly (`test_not_does_not_leak_env`), through a relational rule (`test_not_in_relation_does_not_leak_env`), and through the YAML rule path (`test_not_does_not_leak_env_via_yaml`).

## Test plan

- [x] All three new regression tests fail before the fix and pass after.
- [x] `cargo test -p ast-grep-core -p ast-grep-config` all green (no regressions).
- [x] `cargo clippy -p ast-grep-core -p ast-grep-config --all-targets` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where negated pattern matches could leak temporary variable bindings into the surrounding match environment; negation now isolates probe bindings so subsequent matches behave correctly.

* **Tests**
  * Added regression tests that verify negated rules do not persist internal bindings across matching scenarios.

### DIFF
--- a/crates/config/src/rule/relational_rule.rs
+++ b/crates/config/src/rule/relational_rule.rs
@@ -337,6 +337,39 @@ mod test {
     o::All::new(vec![Rule::Pattern(Pattern::new(target, TS::Tsx)), relation])
   }
 
+  // A `not` sub-rule must never export a metavariable binding. When the negated
+  // inner matches some candidate (making the negation fail) the relational
+  // `find_map` keeps probing later candidates with the SAME env; without
+  // rollback in `Not`, the inner's stray binding survives into the eventual
+  // successful match. Here the nearest preceding sibling is a `return $A`
+  // (negated inner matches, would bind A=foo), while a farther sibling lets the
+  // negation succeed, so the match should bind nothing.
+  #[test]
+  fn test_not_in_relation_does_not_leak_env() {
+    use ast_grep_core::ops as o;
+    let follows = Follows {
+      former: Rule::Not(Box::new(o::Not::new(Rule::Pattern(Pattern::new(
+        "return $A",
+        TS::Tsx,
+      ))))),
+      stop_by: StopBy::End,
+    };
+    let grep = TS::Tsx.ast_grep("function f() { bar(); return foo; target; }");
+    let root = grep.root();
+    let target = root
+      .dfs()
+      .find(|n| n.text() == "target;")
+      .expect("find target stmt");
+    let mut env = Cow::Owned(MetaVarEnv::new());
+    let matched = follows.match_node_with_env(target, &mut env);
+    assert!(matched.is_some(), "target follows a non-return statement");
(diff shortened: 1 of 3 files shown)


knutwannheden:
Heads-up: the failing `coverage` check here is **not** a coverage regression — it's the Codecov **upload** step failing to GPG-verify the Codecov CLI binary it downloads:

```
gpg: Can't check signature: No public key
==> Could not verify signature. Please contact Codecov if problem continues
    Exiting...
##[error]Process completed with exit code 1.
```

Since `.github/workflows/coverage.yaml` sets `fail_ci_if_error: true`, this transient download/verification flake fails the whole job. It isn't specific to this PR:

- this same branch passed the `coverage` job on one run and failed on two others, and
- the sibling PR #2677 passed `coverage`.

The patch coverage itself is fine — I verified locally with `cargo llvm-cov` that the changed lines are fully covered. A re-run of the job should turn it green.

HerringtonDarkholme (approved):

PR #2679, opened 2026-06-02
fix/dart-expression-patterns

asmrtfm:
# fix: support expression-level patterns for Dart language

## Changes
* Dart's tree-sitter grammar only accepts declarations at the top level, causing expression patterns like `print($A)` and statement patterns like `if ($COND) { $$$BODY }` to be misinterpreted as function declarations.
* This adds a custom `build_pattern` impl for Dart that detects misparses (ERROR/MISSING nodes, statement keywords) and wraps the pattern inside a function body to get a correct parse. A post-processing step fixes the Dart grammar ambiguity where `{ $$$BODY }` is parsed as a set_or_map_literal instead of a block.
* Also fixes SgLang::build_pattern in the CLI to delegate to the underlying language impl instead of bypassing custom build_pattern implementations.

## Summary
  - Add custom `build_pattern` for Dart that wraps expression/statement patterns in a function body, fixing misparses caused by Dart's declaration-only top-level grammar
  - Fix `SgLang::build_pattern` to delegate to underlying language impls instead of bypassing them
  - Handle Dart's `{ $$$BODY }` set_or_map_literal vs block ambiguity via pattern tree post-processing

## Test plan
  - 14 Dart-specific tests covering declarations, expressions, statements, rewrites
  - Full suite (645+ tests) passes with 0 failures
  - Verified against real Flutter project: `if`, `setState`, `StatefulWidget` patterns all match
  - `cargo fmt`, `cargo clippy -D clippy::all` clean

## Summary by CodeRabbit

* **New Features**
  * Enhanced Dart language support for matching expressions and statements with improved parsing capabilities.

* **Tests**
  * Added comprehensive test coverage for Dart expression and statement matching.

* **Chores**
  * Internal refactoring to support language-specific pattern implementations.

### DIFF
--- a/crates/cli/src/lang/mod.rs
+++ b/crates/cli/src/lang/mod.rs
@@ -198,7 +198,10 @@ impl Language for SgLang {
       .or_else(|| SupportLang::from_path(path).map(Builtin))
   }
   fn build_pattern(&self, builder: &PatternBuilder) -> std::result::Result<Pattern, PatternError> {
-    builder.build(|src| StrDoc::try_new(src, *self))
+    match self {
+      Builtin(b) => b.build_pattern(builder),
+      Custom(c) => c.build_pattern(builder),
+    }
   }
 }
 
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -19,7 +19,8 @@ pub mod tree_sitter;
 pub mod pinned;
 
 mod match_tree;
-mod node;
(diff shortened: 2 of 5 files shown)


HerringtonDarkholme:
ast-grep does not support per-language custom parsing. If you need better language support please try language specific tooling. 

Otherwise, try pattern object https://ast-grep.github.io/guide/rule-config/atomic-rule.html#pattern-object or `kind`.

Alternatively, asking upstream tree-sitter maintainer to support expression level parsing is another solution

PR #2690, opened 2026-06-07
feat: add outline item model

HerringtonDarkholme:
## Summary
- add the internal outline data model: files, items, positions, ranges, symbol types, and roles
- represent roles as a de-duplicated multi-role set so one item can be both definition/export/import
- add focused unit tests for role/type filtering and multi-role behavior

## Notes
- extraction and rendering intentionally remain out of scope for this PR
- the model module has a scoped dead-code allowance until later slices consume it

## Test
- cargo test -p ast-grep outline::model --lib
- commit hook: fmt, cargo check, clippy

## Summary by CodeRabbit

* **Refactor**
  * Simplified the internal outline representation and improved handling of multi-role outline symbols; no change to CLI behavior or public outputs.
* **Chores**
  * Added and updated tests to validate multi-role behavior, flag uniqueness, and role-matching semantics.

### DIFF
--- a/crates/cli/src/outline/model.rs
+++ b/crates/cli/src/outline/model.rs
@@ -1,0 +1,195 @@
+use std::ops::Range;
+
+use ast_grep_core::Position;
+
+/// Outline symbol category.
+///
+/// The names follow LSP `DocumentSymbol.kind`, but ast-grep stores the symbolic
+/// category directly instead of exposing LSP numeric values.
+/// See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_documentSymbol
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SymbolType {
+  File,
+  Module,
+  Namespace,
+  Package,
+  Class,
+  Method,
+  Property,
+  Field,
+  Constructor,
+  Enum,
+  Interface,
+  Function,
+  Variable,
+  Constant,
+  String,
+  Number,
+  Boolean,
+  Array,
+  Object,
+  Key,
+  Null,
+  EnumMember,
+  Struct,
+  Event,
+  Operator,
+  TypeParameter,
+}
+
+/// A file-level relationship between an outline item and its source file.
+///
+/// Roles are facets, not mutually exclusive categories.
+/// One item can be both a definition and an export.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum OutlineRole {
+  /// A symbol declared or defined in the file.
+  Definition,
+  /// A symbol or module brought into the file from elsewhere.
+  Import,
+  /// A symbol or module exposed as part of the file's outward surface.
+  Export,
+}
+
+/// Compact role flags for one outline item.
+///
+/// A single item can have multiple roles.
+/// e.g. `pub struct Foo` is both a definition and an export.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct OutlineRoles(u8);
+
+impl OutlineRoles {
+  const DEFINITION: u8 = 1 << 0;
+  const IMPORT: u8 = 1 << 1;
+  const EXPORT: u8 = 1 << 2;
+
+  /// Creates role flags from zero or more roles.
+  pub fn new(roles: impl IntoIterator<Item = OutlineRole>) -> Self {
+    roles
+      .into_iter()
+      .fold(Self::default(), |roles, role| roles.with(role))
+  }
+
+  /// Returns true when the role flag is present.
+  pub fn contains(self, role: OutlineRole) -> bool {
+    self.0 & Self::flag(role) != 0
+  }
+
+  /// Returns true when all role flags in `roles` are present.
+  pub fn contains_all(self, roles: OutlineRoles) -> bool {
+    self.0 & roles.0 == roles.0
+  }
+
+  /// Returns true when no role flag is present.
+  pub fn is_empty(self) -> bool {
+    self.0 == 0
+  }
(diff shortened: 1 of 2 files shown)


crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  EnumMember,
+  Struct,
+  Event,
+  Operator,
+  TypeParameter,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum OutlineRole {
```
HerringtonDarkholme, line 45: please add comment explain the design, use triple slash

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
+use std::collections::BTreeSet;
+use std::path::PathBuf;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum SymbolType {
```
HerringtonDarkholme, line 11: add comment how symbolType is designed, and its origin from LSP protocol.  

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_documentSymbol

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+pub enum OutlineRole {
+  Definition,
+  Import,
+  Export,
+}
+
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct OutlineRoles {
+  roles: BTreeSet<OutlineRole>,
```
HerringtonDarkholme: do you really need btree set? it can be a simple u8 to represent these flags

also, add public triple slash comments

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  pub fn intersects(&self, roles: &OutlineRoles) -> bool {
+    self.roles.iter().any(|role| roles.contains(*role))
+  }
+
+  pub fn is_empty(&self) -> bool {
+    self.roles.is_empty()
+  }
+}
+
+impl<const N: usize> From<[OutlineRole; N]> for OutlineRoles {
```
HerringtonDarkholme: wrong abstraction

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  fn from(roles: [OutlineRole; N]) -> Self {
+    Self::new(roles)
+  }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OutlinePosition {
```
HerringtonDarkholme: can you try reusing the repo's existing stuff?

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+      line,
+      column,
+      byte_offset,
+    }
+  }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OutlineRange {
```
HerringtonDarkholme: ditto

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+
+impl OutlineRange {
+  pub fn new(start: OutlinePosition, end: OutlinePosition) -> Self {
+    Self { start, end }
+  }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OutlineItem {
```
HerringtonDarkholme: this item does not need Serialize so probably you don't need them, so roles can be a u8 to save some allocation

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  pub name: String,
```
HerringtonDarkholme: can you use &str for name? the string source can come from tree-sitter node

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  pub symbol_type: SymbolType,
+  pub roles: OutlineRoles,
+  pub range: OutlineRange,
+  pub signature: String,
```
HerringtonDarkholme: how do you know signature of an item? how does it interact with the rule extraction?

not all symbols have signature.

crates/cli/src/outline/model.rs
```diff
@@ -0,0 +1,281 @@
...
+  pub fn empty(path: PathBuf) -> Self {
+    Self {
+      path,
+      items: vec![],
+    }
+  }
+}
+
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct OutlineFilter {
```
HerringtonDarkholme: remvoe this, think it later. we have not encoutnered the filtering part of this feature.

crates/cli/src/outline/model.rs
HerringtonDarkholme: please also add specification link

crates/cli/src/outline/model.rs
HerringtonDarkholme: you didn't explain what is definition/export/import. add them

PR #2691, opened 2026-06-07
feat: add outline rule definition

HerringtonDarkholme:
## Summary
- add the serialized outline extractor contract backed by ast-grep `SerializableRuleCore`
- define name sources for matched node text and ast-grep meta-variable captures
- normalize extractor roles into compact `OutlineRoles` flags and reject empty role sets

## Notes
- no builtin outline rules are added in this PR
- no extraction runtime or rendering behavior is added in this PR
- `SymbolType` and `OutlineRole` now derive `Deserialize` so extractor definitions can use their existing model values

## Test
- cargo test -p ast-grep outline::rule --lib
- cargo test -p ast-grep outline::model --lib
- cargo test -p ast-grep test_outline --lib
- cargo clippy -p ast-grep --lib --tests
- commit hook: fmt, cargo check, clippy

## Summary by CodeRabbit

* **New Features**
  * Added an outline rules framework for defining and normalizing extractor rules with role validation.
  * Configuration now accepts camelCase names for symbol types and roles.
* **Bug Fixes**
  * Rejects rules with empty role sets to prevent invalid extractor configurations.
* **Tests**
  * Added unit tests validating YAML parsing, normalization behavior, and failure on empty roles.

### DIFF
--- a/crates/cli/src/outline/rule.rs
+++ b/crates/cli/src/outline/rule.rs
@@ -1,0 +1,178 @@
+use std::fmt::{Display, Formatter};
+
+use ast_grep_config::SerializableRuleCore;
+use serde::Deserialize;
+
+use crate::lang::SgLang;
+use crate::outline::model::{OutlineRole, OutlineRoles, SymbolType};
+
+/// Serialized outline extractor definition.
+///
+/// The `rule`, `constraints`, `utils`, and `transform` fields are flattened
+/// from ast-grep's existing rule core. Outline extraction reuses ast-grep rules
+/// instead of adding a tree-sitter query format.
+///
+/// The matched node's source range becomes the outline item range. The `name`
+/// field is an ast-grep template string used to derive the display name from
+/// the match. For example, `"$NAME"` can reference a meta-variable captured by
+/// the rule.
+#[derive(Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SerializableOutlineRule {
+  /// Unique, descriptive identifier for this extractor.
+  id: String,
+  /// Language this extractor applies to.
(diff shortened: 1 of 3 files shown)


crates/cli/src/outline/rule.rs
```diff
@@ -0,0 +1,184 @@
...
+  /// Unique, descriptive identifier for this extractor.
+  pub id: String,
+  /// Language this extractor applies to.
+  pub language: SgLang,
+  /// Symbol category assigned to matches from this extractor.
+  pub symbol_type: SymbolType,
+  /// File-level roles assigned to matches from this extractor.
+  pub roles: OutlineRoles,
+  /// How to compute the outline item name from a match.
+  pub name: OutlineNameSource,
```
HerringtonDarkholme: I don't think we need a new OutlineNameSource, we can reuse ast-grep's TemplateFix?

crates/cli/src/outline/rule.rs
```diff
@@ -0,0 +1,184 @@
...
+  /// Unique, descriptive identifier for this extractor.
+  pub id: String,
+  /// Language this extractor applies to.
+  pub language: SgLang,
+  /// Symbol category assigned to matches from this extractor.
+  pub symbol_type: SymbolType,
+  /// File-level roles assigned to matches from this extractor.
+  pub roles: Vec<OutlineRole>,
+  /// How to compute the outline item name from a match.
+  pub name: OutlineNameSource,
```
HerringtonDarkholme: This can be just String

aeroxy:
Great, I'll reuse some of these features downstream! <3

One less moving part to maintain.

PR #2700, opened 2026-06-11
feat: add outline CLI shell

HerringtonDarkholme:
## Summary
- add the public `sg outline` CLI shell and documented options
- wire outline to shared input/color/json CLI argument conventions
- keep runtime behavior as a placeholder `nothing found` for this slice
- expand CLI parser coverage for the outline command contract

## Verification
- `cargo fmt --check`
- `cargo test -p ast-grep --lib test_cli`
- `cargo check -p ast-grep`
- `cargo clippy -p ast-grep --all-targets -- -D warnings`

## Summary by CodeRabbit

* **New Features**
  * Expanded outline CLI options: language selection, multiple JSON formats (compact/stream), color control, item/type filtering, name matching, public-member filtering, multiple view modes, stdin/language handling, follow/no-ignore/globs/threads.

* **Tests**
  * Broadened test matrix covering many valid and invalid outline argument combinations and validations.

* **Bug Fixes**
  * Standardized output when no outline items are found (now prints "nothing found").

### DIFF
--- a/crates/cli/src/outline.rs
+++ b/crates/cli/src/outline.rs
@@ -1,26 +1,125 @@
 // Extraction and rendering will consume this model in later slices.
 #[allow(dead_code)]
 mod model;
-// Builtin and custom extractors will consume this rule contract later.
+// Builtin extractors will consume this rule contract later.
 #[allow(dead_code)]
 mod rule;
 
-use std::path::PathBuf;
 use std::process::ExitCode;
 
-use clap::Args;
+use clap::{Args, ValueEnum};
+
+use crate::lang::SgLang;
+use crate::print::{ColorArg, JsonStyle};
+use crate::utils::InputArgs;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
+enum OutlineItems {
+  /// Use `structure` for file or stdin input, `exports` when any directory is given.
+  Auto,
+  /// Top-level items defined locally in the file, excluding imports.
+  Structure,
+  /// Top-level items exported from the file or module.
+  Exports,
+  /// Top-level items imported from other files or modules.
+  Imports,
+  /// All top-level items, including imports and exports.
+  All,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, ValueEnum)]
+enum OutlineView {
+  /// Use `digest` for file or stdin input, `names` when any directory is given.
+  Auto,
+  /// One grouped name line per symbol type for each file.
+  Names,
+  /// One source/signature line per top-level item.
+  Signatures,
+  /// Signatures plus compact member name digests.
+  Digest,
+  /// Signatures plus one source/signature line per direct member.
+  Expanded,
+}
 
 #[derive(Args)]
 pub struct OutlineArg {
-  /// The files or directories to summarize.
+  /// Specify the input language.
   ///
-  /// Accepts one or more file/directory paths.
-  #[clap(value_name = "PATH", required = true)]
-  paths: Vec<PathBuf>,
+  /// For path input, ast-grep parses only files of this language. For stdin,
+  /// this flag is required because there is no file path to infer the language from.
+  #[clap(short, long, required_if_eq("stdin", "true"))]
+  lang: Option<SgLang>,
+
+  /// Output outline entries in structured JSON.
+  ///
+  /// If this flag is set, ast-grep will output outline entries in JSON format.
+  /// You can pass optional value to this flag by using `--json=<STYLE>` syntax
+  /// to further control how JSON object is formatted and printed. ast-grep will
+  /// `pretty`-print JSON if no value is passed.
+  /// Note, the json flag must use `=` to specify its value.
+  #[clap(
+      long,
+      value_name="STYLE",
+      num_args(0..=1),
+      require_equals = true,
+      default_missing_value = "pretty"
+  )]
+  json: Option<JsonStyle>,
+
+  /// Controls output color.
+  ///
+  /// This flag controls when to use colors. The default setting is 'auto', which
+  /// means ast-grep will try to guess when to use colors. If ast-grep is
+  /// printing to a terminal, then it will use colors, but if it is redirected to a
+  /// file or a pipe, then it will suppress color output. ast-grep will also suppress
+  /// color output in some other circumstances. For example, no color will be used
+  /// if the TERM environment variable is not set or set to 'dumb'.
+  #[clap(long, default_value = "auto", value_name = "WHEN")]
+  color: ColorArg,
+
+  /// Select which top-level items to include.
+  ///
+  /// This option controls top-level structure such as classes, structs, interfaces,
+  /// functions, and modules. It does not filter members.
+  /// By default, ast-grep picks the items automatically based on the input path.
+  #[clap(long, default_value = "auto", value_name = "ITEMS")]
+  items: OutlineItems,
+
+  /// Keep only top-level items with these comma-separated symbol types.
+  ///
+  /// For example, `--type class,enum` keeps both classes and enums.
+  #[clap(long = "type", value_name = "TYPE[,TYPE...]")]
+  symbol_type: Option<String>,
+
+  /// Keep only top-level items matching this regex.
+  ///
+  /// The regex is matched against item names, signatures, first source lines,
+  /// and import/export targets and aliases. It never matches members.
+  #[clap(long = "match", value_name = "REGEX")]
+  match_item: Option<String>,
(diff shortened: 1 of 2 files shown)


crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  Names,
+  Signatures,
+  Digest,
+  Expanded,
+}
 
 #[derive(Args)]
 pub struct OutlineArg {
-  /// The files or directories to summarize.
+  /// Parse input as a specific language.
```
HerringtonDarkholme: wrong explanation, --lang is also used for filtering the file to parse. so this should be something like specify the input language

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
   ///
-  /// Accepts one or more file/directory paths.
-  #[clap(value_name = "PATH", required = true)]
-  paths: Vec<PathBuf>,
+  /// This is required when reading from standard input. For path input,
+  /// ast-grep infers the language from each file path unless this option is set.
+  #[clap(short, long, required_if_eq("stdin", "true"))]
+  lang: Option<SgLang>,
+
+  /// Output outline entries as JSON.
```
HerringtonDarkholme: write something like --json in args.rs

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(
+      long,
+      value_name="STYLE",
+      num_args(0..=1),
+      require_equals = true,
+      default_missing_value = "pretty"
+  )]
+  json: Option<JsonStyle>,
+
+  /// Controls output color in text mode.
```
HerringtonDarkholme: ditto, write something similar to --color in args

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(long, default_value = "auto", value_name = "ITEMS")]
+  items: OutlineItems,
+
+  /// Keep only top-level items with these comma-separated symbol types.
+  ///
+  /// Examples: `function`, `struct,enum`, `class,interface`.
+  #[clap(long = "type", value_name = "TYPE[,TYPE...]")]
+  symbol_type: Option<String>,
+
+  /// Keep only top-level items whose useful fields match this regex.
```
HerringtonDarkholme: i think this should be changed to "name" matches this regex. do not make it match fields

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(long = "match", value_name = "REGEX")]
+  match_item: Option<String>,
+
+  /// Display only public members in member views.
```
HerringtonDarkholme, line 101: also add what's the default behavior if this flag is not set

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(long)]
+  pub_members: bool,
+
+  /// Select the text presentation.
```
HerringtonDarkholme, line 108: explain what's text presentation

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(long, default_value = "auto", value_name = "VIEW")]
+  view: OutlineView,
+
+  /// Load additional outline extractor definitions.
+  #[clap(long, value_name = "FILE", action = clap::ArgAction::Append)]
+  outline_rules: Vec<PathBuf>,
```
HerringtonDarkholme: remove this for now, i have not considered how to do custom stuff now

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+
+  /// Disable bundled outline extractor definitions.
+  #[clap(long)]
+  no_default_outline_rules: bool,
```
HerringtonDarkholme: ditto remove this for now, i have not considered how to do custom stuff now

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+  #[clap(long, default_value = "auto", value_name = "WHEN")]
+  color: ColorArg,
+
+  /// Select which top-level items to include.
+  #[clap(long, default_value = "auto", value_name = "ITEMS")]
+  items: OutlineItems,
+
+  /// Keep only top-level items with these comma-separated symbol types.
+  ///
+  /// Examples: `function`, `struct,enum`, `class,interface`.
```
HerringtonDarkholme: explain if this is `and` or `or` logic

crates/cli/src/outline.rs
```diff
@@ -8,19 +8,94 @@ mod rule;
...
+      require_equals = true,
+      default_missing_value = "pretty"
+  )]
+  json: Option<JsonStyle>,
+
+  /// Controls output color in text mode.
+  #[clap(long, default_value = "auto", value_name = "WHEN")]
+  color: ColorArg,
+
+  /// Select which top-level items to include.
```
HerringtonDarkholme, line 80: explain more. also add that items does not filter member

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,107 @@
...
+  Expanded,
+}
 
 #[derive(Args)]
 pub struct OutlineArg {
-  /// The files or directories to summarize.
+  /// Specify the input language.
+  ///
+  /// For path input, ast-grep parses only files of this language. For stdin,
+  /// this flag is required because there is no file path for language inference.
```
HerringtonDarkholme: `for stdin, this flag is required because there is no file path for language inference.`

this sentence is too weak. change it to something `this flag is required for language inference without file path` don't add "because, there is" weak/unnecessary stuff

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,107 @@
...
+  #[clap(long = "type", value_name = "TYPE[,TYPE...]")]
+  symbol_type: Option<String>,
+
+  /// Keep only top-level items whose name matches this regex.
+  #[clap(long = "match", value_name = "REGEX")]
+  match_item: Option<String>,
+
+  /// Display only public members in member views.
+  ///
+  /// By default, member views display all extracted members.
```
HerringtonDarkholme: add that public member will be sorted before non-public mebmer

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,114 @@
...
+  /// printing to a terminal, then it will use colors, but if it is redirected to a
+  /// file or a pipe, then it will suppress color output. ast-grep will also suppress
+  /// color output in some other circumstances. For example, no color will be used
+  /// if the TERM environment variable is not set or set to 'dumb'.
+  #[clap(long, default_value = "auto", value_name = "WHEN")]
+  color: ColorArg,
+
+  /// Select which top-level items to include.
+  ///
+  /// This option controls file-level structure such as definitions, imports, and
```
HerringtonDarkholme: do not use terms like `definitions, imports, and exports`. it is superseded. 

Use concrete examples like `class, struct, interface`. file-level can also be changed to `top-level`

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,114 @@
...
+  /// exports. It does not filter members; member output is controlled by --view
```
HerringtonDarkholme: no need to add `member output is controlled by `

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,114 @@
...
+  /// and --pub-members.
+  #[clap(long, default_value = "auto", value_name = "ITEMS")]
+  items: OutlineItems,
+
+  /// Keep only top-level items with these comma-separated symbol types.
+  ///
+  /// Multiple types use OR logic: `--type struct,enum` keeps struct or enum
```
HerringtonDarkholme: the sentence `OR` logic sounds not like help text, also one example can explain it, no need to add examples as separate line

crates/cli/src/outline.rs
```diff
@@ -1,26 +1,114 @@
...
-  /// Accepts one or more file/directory paths.
-  #[clap(value_name = "PATH", required = true)]
-  paths: Vec<PathBuf>,
+  /// By default, member views display all extracted members.
+  #[clap(long)]
+  pub_members: bool,
+
+  /// Select the text presentation.
+  ///
+  /// This controls how much source structure is printed in text output:
```
HerringtonDarkholme: add that view option has increasingly more information

PR #2745, opened 2026-06-19
feat: add outline extractor options

HerringtonDarkholme:
Stacked on #2744.\n\nThis PR introduces core outline extractor options for item/member filters and entry detail without wiring them into extraction yet.

## Summary by CodeRabbit

* **New Features**
  * Added a new public `options` module for the outline crate.
  * Introduced configurable outline extraction options, including symbol-type allowlists, import/export filtering, and output detail selection (name vs. signature).
  * Added member-level options to optionally include member details, including configurable public-member filtering.
* **Documentation**
  * Exposed the new options through the public API so callers can customize outline generation behavior.

PR #2746, opened 2026-06-19
feat: apply outline extractor options

HerringtonDarkholme:
Stacked on #2745.\n\nThis PR wires outline extractor options into rule compilation and extraction, including safe rule pruning, runtime filtering, and name-only signature detail.

## Release Notes

* **New Features**
  * Added configurable extraction options to control whether items and members are compiled and retained during outline processing.
  * Introduced an entry “detail” mode to switch output between name-only and full signature rendering.

* **Enhancements**
  * Extraction now applies option-driven filtering consistently, so only matching rules are compiled and extracted (including correct import/export classification).
  * Items and members are included or suppressed according to the selected retention settings.

* **Tests**
  * Updated and expanded unit tests to cover member/item suppression, detail-driven signature behavior, and import/export filtering.

### DIFF
--- a/crates/outline/src/combined_extractor.rs
+++ b/crates/outline/src/combined_extractor.rs
@@ -23,6 +23,7 @@ use std::collections::HashMap;
 
 use crate::extractor::{ItemExtractor, MemberExtractor, OutlineRuleError, SerializableOutlineRule};
 use crate::model::{OutlineItem, OutlineMember};
+use crate::options::OutlineExtractorOptions;
 
 /// Runtime outline extractors organized for a shared item traversal.
 pub struct CombinedExtractors<L: Language> {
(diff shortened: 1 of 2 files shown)


crates/outline/src/combined_extractor.rs
```diff
@@ -53,25 +56,62 @@ impl<L: Language> CombinedExtractors<L> {
   pub fn try_from(
     extractors: Vec<SerializableOutlineRule<L>>,
     globals: &GlobalRules,
+  ) -> Result<Self, OutlineRuleError> {
+    Self::try_from_rules(extractors, &OutlineExtractorOptions::default(), globals)
+  }
+
+  pub fn try_from_rules(
+    extractors: Vec<SerializableOutlineRule<L>>,
+    options: &OutlineExtractorOptions,
```
HerringtonDarkholme: this can take the options Ownership

PR #2751, opened 2026-06-21
perf: use single traversal for outline extraction

HerringtonDarkholme:
## Summary
- add a cursor-backed pruning traversal for outline extraction
- run item extraction and scoped member extraction in one AST traversal
- avoid visiting descendants once an item/member match owns that subtree

## Benchmark context
In the commit-impact matrix before splitting PRs, this was the second-highest impact change after compiled transform regexes:
- traversal-only geomean: 1.11x vs baseline
- per-view traversal-only: names 1.06x, signatures 1.15x, expanded 1.07x, digest 1.16x

## Test
- cargo test -p ast-grep-core test_prune_pre_order_skips_subtree
- cargo test -p ast-grep-outline combined_extractor --lib
- cargo test -p ast-grep-outline -p ast-grep --lib outline
- cargo clippy -p ast-grep-core -p ast-grep-outline --lib --all-targets

## Summary by CodeRabbit

* **New Features**
  * Added a caller-controlled pre-order tree traversal cursor with subtree pruning and a scoping handle to detect when traversal exits a captured subtree.
* **Refactor**
  * Updated outline extraction to use a single traversal with explicit item vs. member scope handling, improving how matched items and their scoped members are collected.
* **Tests**
  * Added coverage to ensure pruned subtrees are skipped as expected and that matching correctly resumes after completing a member scope.

### DIFF
--- a/crates/core/src/tree_sitter/traversal.rs
+++ b/crates/core/src/tree_sitter/traversal.rs
@@ -285,6 +285,84 @@ impl<'t, L: LanguageExt> Traversal<'t, StrDoc<L>> for Pre<'t, L> {
   }
 }
 
+/// Pre-order cursor traversal where the caller decides whether to enter children.
+///
+/// This is useful when matching the current node determines whether its whole
+/// subtree can be skipped. Unlike [`Pre`] plus reentrancy calibration, this
+/// traversal does not step into a child before the caller has made that choice.
+pub struct Prune<'tree, L: LanguageExt> {
+  cursor: ts::TreeCursor<'tree>,
+  root: &'tree Root<StrDoc<L>>,
+  start_id: Option<usize>,
+  current_depth: usize,
+}
+
+/// Opaque marker for a subtree in a [`Prune`] traversal.
+///
+/// Callers can store this when visiting a node and later ask whether traversal
+/// has moved past that node's subtree without depending on cursor depth.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PruneSubtree<'tree> {
+  root_id: usize,
+  root_depth: usize,
+  tree: PhantomData<&'tree ()>,
+}
+
+impl<'tree, L: LanguageExt> Prune<'tree, L> {
+  pub fn new(node: &Node<'tree, StrDoc<L>>) -> Self {
+    Self {
+      cursor: node.inner.walk(),
+      root: node.root,
+      start_id: Some(node.inner.id()),
+      current_depth: 0,
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


crates/cli/src/outline/extract.rs
```diff
@@ -220,8 +220,15 @@ fn extract_path(
   let Some(lang) = lang.or_else(|| SgLang::from_path(path)) else {
     return Ok(None);
   };
-  let source =
-    read_file(path).with_context(|| format!("Cannot extract outline from {}", path.display()))?;
+  let source = match read_file(path) {
+    Ok(source) => source,
+    Err(err) if err.to_string() == "File is empty" => {
+      return Ok(empty_outline_file(path, lang, extractors.show_empty_files));
```
HerringtonDarkholme: do you need add this? is it related to perf?

crates/cli/src/outline/extract.rs
```diff
@@ -234,6 +241,14 @@ fn extract_path(
   }))
 }
 
+fn empty_outline_file(path: &Path, lang: SgLang, show: bool) -> Option<OutlineFile<'static>> {
```
HerringtonDarkholme: ditto

crates/core/src/tree_sitter/traversal.rs
```diff
@@ -285,6 +285,61 @@ impl<'t, L: LanguageExt> Traversal<'t, StrDoc<L>> for Pre<'t, L> {
...
+      start_id: Some(node.inner.id()),
+      current_depth: 0,
+    }
+  }
+
+  pub fn current(&self) -> Option<Node<'tree, StrDoc<L>>> {
+    self.start_id.map(|_| self.root.adopt(self.cursor.node()))
+  }
+
+  pub fn current_depth(&self) -> usize {
```
HerringtonDarkholme: current and current_depth are too similar in naming but differnt, at least changing current to current_node

crates/cli/src/outline/extract.rs
HerringtonDarkholme: Removed this unrelated empty-file handling from the perf PR.

crates/cli/src/outline/extract.rs
HerringtonDarkholme: Removed the helper and test together with the unrelated empty-file handling.

crates/core/src/tree_sitter/traversal.rs
HerringtonDarkholme: Renamed Prune::current to Prune::current_node and updated call sites.

PR #2752, opened 2026-06-21
perf: lazily render outline transforms

HerringtonDarkholme:
Summary:
- Parse outline transforms only when name/signature templates reference transform vars.
- Apply transforms lazily after a node match during extraction.
- Keep member traversal state inline with the PruneSubtree scope API.

Tests:
- cargo test -p ast-grep-outline --lib
- cargo test -p ast-grep-outline -p ast-grep --lib outline
- cargo test -p ast-grep-config --lib transform
- cargo clippy -p ast-grep-config -p ast-grep-outline --lib --all-targets

## Summary by CodeRabbit

* **New Features**
  * Added support for applying transforms to outline entry extraction with validation for undefined rewriters.

* **Refactor**
  * Restructured internal outline extraction traversal logic for improved architecture.
  * Enhanced rule processing with new transform validation and application methods.

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -29,7 +29,7 @@ pub use rule_config::{
   Metadata, RuleConfig, RuleConfigError, SerializableRewriter, SerializableRuleConfig, Severity,
 };
 pub use rule_core::{RuleCore, RuleCoreError, SerializableRuleCore};
-pub use transform::Transformation;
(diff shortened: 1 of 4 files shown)


HerringtonDarkholme:
Closing this approach. Lazy outline transform rendering leaks too much config/transform abstraction into outline extraction, and the dependency graph case is unsafe: a template may reference one transform var that depends on another transform var. Keeping eager transform handling is the better tradeoff here.

PR #2806, opened 2026-07-14
fix(core): cap display_context walk distance to avoid output blowup on minified files

BezSaharaD:
## Summary

Minified/bundled source files routinely have a single "line" spanning the entire file (no `\n` at all). `display_context` (used to build the `lines` field shown around every match, both in `--json` output and the terminal printer) walks byte-by-byte from a match looking for line boundaries with no upper bound on how far it walks — so on such files this walk degenerates to O(file size) *per match*, and a file with many matches on that one line (a very common real-world case for tools that scan `node_modules`/vendored bundles) produces `lines` context output that is O(matches × file size).

### Repro

A synthetic 213KB single-line file with 5000 `instanceof` matches produced a **1.07GB** `--json=stream` report — each of the 5000 match objects embeds the *entire* 213KB line verbatim in its `lines` field (a >5000x amplification from input to output size).

## Fix (first commit)

Cap how far the leading/trailing walk goes in `display_context`, scaled by the requested `before`/`after` line counts (1000 bytes per requested line by default) so legitimate multi-line `-A`/`-B`/`-C` context on normal code isn't cut short. Added a char-boundary nudge after the capped walk, since an arbitrary byte-distance cutoff (unlike the existing newline-based stopping conditions) can land mid-codepoint.

With the fix, the same synthetic repro produces **14.4MB** (74x smaller) with identical match count and content; a normal short-line file's `lines` output is byte-for-byte unchanged (well under the cap).

## New flag (second commit)

Added `--max-context-bytes NUM` (flows through `ContextArgs`, both output paths) so the cap is user-configurable instead of a fixed constant — default matches the previous hardcoded value. Verified it scales monotonically: 50 → 4.87MB, default (1000) → 14.37MB, 5000 → 53.79MB on the same synthetic repro.

## Testing

- `cargo test -p ast-grep-core` (154 tests) and `cargo test -p ast-grep` (full CLI suite) — all pass, including the existing `test_display_context`/`test_multi_line_context`.
- Ran our own project's full custom ast-grep rule set (46 rules) against the **entire `node_modules`** of a real monorepo (1126 packages, ~1.2GB, no filtering of minified/bundled files needed) as a stress test: completed with **0 crashes**, **161,239 matches** (identical count to a pre-fix run against the same tree, confirming no matches lost), **905MB** output — vs. the pre-fix binary's ~16GB for the identical scan.
- Spot-checked the small number of remaining large `lines` fields in that 905MB output: all had single-digit-or-zero `leading`/`trailing` char counts, with the size coming entirely from a legitimately large *matched* span (e.g. `no-classes` matching a genuinely large class body) — confirming the cap holds and isn't defeated by some other code path.

## Disclosure

This fix (root cause investigation, patch, and verification) was developed with Claude's (Anthropic) assistance — reviewed and tested locally (synthetic repro measurements, full test suite runs, full node_modules stress test) before submitting.

## Summary by CodeRabbit

* **New Features**
  * Added the `--max-context-bytes` option to limit the amount of surrounding text shown for matches.
  * Added support for configuring this limit in colored and JSON output.
  * Applied the setting consistently across search, scan, match, and diff results.

* **Bug Fixes**
  * Improved context rendering for long lines while preventing invalid text slicing and related crashes.
  * Preserved valid UTF-8 boundaries when limiting displayed context.

### DIFF
--- a/crates/cli/src/print/colored_print.rs
+++ b/crates/cli/src/print/colored_print.rs
@@ -61,6 +61,7 @@ pub struct ColoredPrinter<W: WriteColor> {
   styles: PrintStyles,
   heading: Heading,
   context: (u16, u16),
+  max_context_bytes: usize,
 }
 
 impl<W: WriteColor> ColoredPrinter<W> {
@@ -71,6 +72,7 @@ impl<W: WriteColor> ColoredPrinter<W> {
       config: term::Config::default(),
       heading: Heading::Auto,
       context: (0, 0),
+      max_context_bytes: ast_grep_core::tree_sitter::DEFAULT_MAX_CONTEXT_BYTES_PER_LINE,
     }
   }
 
@@ -101,6 +103,11 @@ impl<W: WriteColor> ColoredPrinter<W> {
     self.config.end_context_lines = context.1 as usize;
     self
   }
+
+  pub fn max_context_bytes(mut self, max_context_bytes: usize) -> Self {
+    self.max_context_bytes = max_context_bytes;
+    self
+  }
 }
 
 impl<W: WriteColor> Printer for ColoredPrinter<W> {
@@ -116,6 +123,7 @@ impl<W: WriteColor> Printer for ColoredPrinter<W> {
       styles: self.styles.clone(),
       heading: self.heading,
       context: self.context,
+      max_context_bytes: self.max_context_bytes,
       markdown,
     }
   }
@@ -171,6 +179,7 @@ pub struct ColoredProcessor {
   heading: Heading,
   markdown: Markdown,
   context: (u16, u16),
+  max_context_bytes: usize,
 }
 
 impl ColoredProcessor {
@@ -339,7 +348,7 @@ fn print_matches_with_heading(
   };
   let source = first_match.root().get_text();
 
-  let mut merger = MatchMerger::new(&first_match, printer.context);
+  let mut merger = MatchMerger::new(&first_match, printer.context, printer.max_context_bytes);
 
   let display = merger.display(&first_match);
   let mut ret = display.leading.to_string();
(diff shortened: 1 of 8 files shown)


HerringtonDarkholme:
Please verify you are a real human instead of agent

BezSaharaD:
я реальный человек

BezSaharaD (commented):
I am a real person, how can I prove it?

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/ast-grep/ast-grep/pull/2806)</sub>

HerringtonDarkholme:
Can you tell me the use case of this change? If you are working with minified JS, https://github.com/pionxzh/wakaru is a better choice

BezSaharaD:
Can I write in Russian?

HerringtonDarkholme:
you can use AI to translate

BezSaharaD:
ok

BezSaharaD:
Not trying to analyze minified code on purpose — this came up scanning node_modules with our own custom rules (dependency/security-audit style use case). Some packages ship minified dist files, ast-grep matches fine in them, but the lines field in --json output balloons because display_context has no bound on how far it walks looking for a line break, and minified files are often one giant line. A single run over our node_modules produced ~16GB of JSON output from ~1.2GB of input purely because of this. wakaru solves a different problem (turning bundled/minified code back into readable source) — it wouldn't help here since the goal isn't reading/reconstructing those files, just getting pattern matches out of a tree that happens to include some without the output blowing up.

HerringtonDarkholme:
I don't think this should be handled in ast-grep core. The `1.07GB --json=stream report` is the final artifact, not ast-grep's memory consumption or peak RSS. 

ast-grep is streaming the result so probably the memory consumption from ast-grep itself is not egregious. The output, as its name implies, is streamed, so it is really the consumer to decide how to handle it correctly. 

Discarding the same line or truncating are both valid options. I don't think this complexity should be added to ast-grep CLI to reduce memory consumption. Instead, consumer already has the power to choose.

PR #2811, opened 2026-07-16
Respect --no-ignore for configured rule directories

### LINKED ISSUE #2803: [bug] --no-ignore does not apply to ruleDirs, silently skipping rules installed in   Python virtualenvs
### Please read the FAQ for the bug you encountered.

- [x] I have read the existing FAQ

### ⏯ Playground Link

N/A - this concerns filesystem rule discovery and cannot be reproduced in the playground.

### 💻 Code

```
tmp=$(mktemp -d) 
cd "$tmp"
git init -q

mkdir -p .venv/rules src 

# uv creates this file inside its virtual environments.
printf '*\n' > .venv/.gitignore

cat > .venv/sgconfig.yml <<'YAML'
ruleDirs:
  - rules
YAML

cat > .venv/rules/no-print.yml <<'YAML'
id: no-print
language: Python
severity: warning
rule:
  pattern: print($$$ARGS)
message: Do not use print
YAML

printf 'print("hello")\n' > src/example.py

ast-grep scan \
  -c .venv/sgconfig.yml \
  --no-ignore=vcs \
  --no-ignore=parent \
  --no-ignore=hidden \
  --inspect=summary \
  src
```

### 🙁 Actual behavior

```
sg: summary|file: scannedFileCount=1,skippedFileCount=0
sg: summary|rule: effectiveRuleCount=0,skippedRuleCount=0
```

No diagnostic is produced. Removing `.venv/.gitignore` immediately changes `effectiveRuleCount` to `1` and reports `no-print`.

This affects Python packages that distribute ast-grep rules. Package managers such as uv install the rules under a virtual environment containing an automatically generated `.gitignore` with `*`. Passing the package’s `sgconfig.yml` with `-c` therefore loads the configuration but silently skips every rule beneath `ruleDirs`.

`--no-ignore` currently affects traversal of source files, but not traversal of configured rule directories.

### 🙂 Expected behavior

An explicitly configured `ruleDirs` path should be discoverable when the corresponding `--no-ignore` options are supplied. At minimum, ast-grep should not silently succeed with zero effective rules when every configured rule is excluded.

The `--no-ignore` settings should be passed to the rule-discovery walker as well as the source-file walker.

### Additional information about the issue

Reproduced with ast-grep 0.44.1.

The source appears to use separate walkers:

  - Source-file walker receives no_ignore
    https://github.com/ast-grep/ast-grep/blob/26f784516b4a9a07f6969b1d5c3020ce42dbf7ed/crates/cli/src/utils/args.rs#L84-L94

  - Rule-directory walker uses default ignore behavior
    https://github.com/ast-grep/ast-grep/blob/26f784516b4a9a07f6969b1d5c3020ce42dbf7ed/crates/cli/src/config.rs#L181-L210

  Related issues:

  - #1625, where the maintainer confirmed ignore configuration was not passed to rule discovery.

  - #628, which specifically discusses sharing rules through PyPI and virtual environments.

Those issues were closed without covering a virtual environment’s internal `.gitignore` file.

tianrking:
Fixes #2803.

Configured rule directories now use the same ignore settings as source traversal, including --no-ignore hidden and --no-ignore vcs. Utility-rule discovery follows the same behavior.

Adds a regression test for rules installed under a Python virtual environment whose internal .gitignore would otherwise hide every configured rule.

Tests: cargo test -p ast-grep --lib test_no_ignore_applies_to_rule_dirs

## Summary by CodeRabbit

* **New Features**
  * Rule discovery now honors the scan’s “no ignore” setting when traversing configured rule directories.
  * The test command now supports `--no-ignore` to include hidden/VCS-related rule directories during discovery.
* **Bug Fixes**
  * Scans and verified test rule discovery no longer miss configured rules when ignore restrictions are disabled.
* **Tests**
  * Added new cases covering rule discovery with ignore handling disabled for both scan and verify flows.

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -1,5 +1,5 @@
 use crate::lang::{CustomLang, LanguageGlobs, SerializableInjection, SgLang};
-use crate::utils::{ErrorContext as EC, RuleOverwrite, RuleTrace};
+use crate::utils::{ErrorContext as EC, NoIgnore, RuleOverwrite, RuleTrace};
 
 use anyhow::{Context, Result};
 use ast_grep_config::{
@@ -88,8 +88,16 @@ impl ProjectConfig {
     &self,
     rule_overwrite: RuleOverwrite,
   ) -> Result<(RuleCollection<SgLang>, RuleTrace)> {
-    let global_rules = find_util_rules(self)?;
-    read_directory_yaml(self, global_rules, rule_overwrite)
+    self.find_rules_with_no_ignore(rule_overwrite, &NoIgnore::default())
+  }
+
+  pub fn find_rules_with_no_ignore(
+    &self,
+    rule_overwrite: RuleOverwrite,
+    no_ignore: &NoIgnore,
+  ) -> Result<(RuleCollection<SgLang>, RuleTrace)> {
+    let global_rules = find_util_rules(self, no_ignore)?;
+    read_directory_yaml(self, global_rules, rule_overwrite, no_ignore)
   }
 
   /// returns a Result of Result.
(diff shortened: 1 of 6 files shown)


HerringtonDarkholme (changes requested):
I would prefer to move NoIgnore to AstGrepConfig

also the change does not cover verify.rs

HerringtonDarkholme:
See https://github.com/ast-grep/ast-grep/issues/2803#issuecomment-5013135215,

the fix here isn't really what ast-grep should do.

PR #2845, opened 2026-07-24
perf(outline): add per-rule stopBy traversal bounds

### LINKED ISSUE #2825: perf(outline): add per-rule stopBy traversal bounds
## Background

After improving parser performance (including testing with the custom optimized tree-sitter build), outline tree traversal became a more visible part of `ast-grep outline` runtime. Profiling identified outline traversal as a meaningful hotspot: the traversal currently considers retained item rules throughout the file tree, and member rules throughout the matched item's subtree, even when a rule can only validly match nodes directly owned by its source node.

The source node should be defined by extractor role:

- item rule: the file's root AST node
- member rule: the enclosing item matched by the parent item extractor

An experiment that marked root-owned TypeScript/TSX export and namespace/container item rules as immediate showed a material improvement on the opencode repository in exports mode. Using the optimized parser build, canonical output remained identical (1,455 files and 7,575 items), while median wall time fell from 174.6 ms to 151.8 ms (13.1%) and user CPU fell from 1.551 s to 1.333 s (14.1%).

A blanket immediate conversion is not correct, however. The TypeScript compiler corpus contains legacy syntax and intentionally malformed/error-recovery fixtures where candidates occur below wrapper nodes. Those cases demonstrated that traversal bounds must be declared per extractor and must coexist correctly with rules that still require deep traversal.

## Requirements

1. Allow both item and member outline rules to be annotated with `stopBy`.
   - Support at least `stopBy: end` and `stopBy: immediate`.
   - Preserve current behavior by defaulting omitted `stopBy` to `end`.
   - For an item rule, evaluate the boundary relative to the file root AST node.
   - For a member rule, evaluate the boundary relative to the enclosing matched item.

2. A rule with `stopBy: immediate` must not participate when visiting non-immediate descendants of its source node.
   - In a mixed rule set, `end` rules must continue to participate below direct children.
   - Rule ordering/precedence at eligible nodes should remain unchanged.

3. If all active rules for a traversal scope are `immediate`, do not descend beyond the source node's direct children.
   - For item traversal, determine this from the retained item rules after outline options have filtered the rule set.
   - For member traversal, determine this from the member rules applicable to the enclosing item's rule ID and the active member options.

## Acceptance criteria

- Add schema/deserialization coverage for item and member `stopBy`, including the default.
- Add traversal tests for immediate-only and mixed immediate/end item rules.
- Add equivalent tests for member rules scoped to an enclosing item.
- Verify that an immediate-only scope prunes descendant traversal rather than merely skipping matcher evaluation.
- Annotate only bundled rules whose matched candidate is directly owned by the role's source node.
- Benchmark `ast-grep outline --items exports` on opencode and confirm output equivalence against `stopBy: end` rules.
- Check the TypeScript corpus separately and document any remaining legacy/error-recovery cases before changing additional bundled rules to `immediate`.

morgan-coded:
# perf(outline): add per-rule stopBy traversal bounds

## Problem I saw

Outline rules currently keep walking to the end of an item or member scope even when a rule can only match the source node's direct children. That makes those rules visit descendants outside their useful boundary.

### Requirements

1. Item and member rules now accept `stopBy: end` and `stopBy: immediate`. Omitting the field defaults to `end`, preserving the existing behavior. Item bounds are relative to the file root; member bounds are relative to the matched item.
2. An `immediate` rule participates at direct children but not at deeper descendants. In mixed scopes, `end` rules continue participating below direct children, and rule precedence is unchanged wherever both rules are eligible.
3. When every active rule in a scope is `immediate`, traversal skips each unmatched direct child's subtree. The item decision is made after outline options retain their active rules; the member decision is made from the active rules for the enclosing item's rule ID.

The traversal partition stays inside `ast-grep-outline`; this does not change `ast-grep-core`.

### Acceptance

1. Deserialization tests cover item and member rules, explicit `end` and `immediate`, omitted-to-`end` defaulting, rejection of an unknown top-level value, and a nested relational `stopBy: neighbor` alongside the new outline field.
2. Item traversal tests cover immediate-only and mixed immediate/end scopes, including direct-child precedence.
3. Member traversal tests cover the equivalent immediate-only and mixed scopes.
4. Test-only visit counters distinguish pruning from matcher filtering. The immediate-only item fixture visits the program and its two direct children, for 3 visits total, while the equivalent `end` traversal visits more nodes. The member fixture also records fewer visits with `immediate` than with `end` while producing the same member output.
5. Forty empirically qualified TypeScript/TSX item rules are now `immediate`:
   - TypeScript: `ts-import`, `ts-export-from`, `ts-export-list`, `ts-export-namespace`, `ts-export-ambient-module`, `ts-ambient-module`, `ts-export-function`, `ts-function`, `ts-export-class`, `ts-class`, `ts-export-interface`, `ts-interface`, `ts-export-type`, `ts-type`, `ts-export-enum`, `ts-enum`, `ts-export-const`, `ts-export-const-typed`, `ts-export-let`, and `ts-export-let-typed`.
   - TSX: `tsx-import`, `tsx-export-from`, `tsx-export-list`, `tsx-export-namespace`, `tsx-export-ambient-module`, `tsx-ambient-module`, `tsx-export-function`, `tsx-function`, `tsx-export-class`, `tsx-class`, `tsx-export-interface`, `tsx-interface`, `tsx-export-type`, `tsx-type`, `tsx-export-enum`, `tsx-enum`, `tsx-export-const`, `tsx-export-const-typed`, `tsx-export-let`, and `tsx-export-let-typed`.
   - `ts-namespace` and `tsx-namespace` remain `end`; all member rules remain `end`. The wrapper-owned `ts-top-level-arrow-function`, `ts-top-level-const`, `ts-top-level-const-typed`, `ts-top-level-let-uninitialized`, `ts-top-level-let`, `ts-top-level-let-typed`, `tsx-top-level-arrow-function`, `tsx-top-level-const`, `tsx-top-level-const-typed`, `tsx-top-level-let-uninitialized`, `tsx-top-level-let`, and `tsx-top-level-let-typed` rules also remain `end`. The 34 nested relational `stopBy: neighbor` values are unchanged.
6. The `sst/opencode` benchmark used commit `909db63265971d67d2fe4ba7f9d7b74cc33e2fdc`, `--items exports`, release mode, one thread, and three alternating wall-clock runs:

   ```text
   end run 1
   real 2.36
   user 2.11
   sys 0.09

   immediate run 1
   real 2.15
   user 2.08
   sys 0.08

   end run 2
   real 2.30
   user 2.14
   sys 0.08

   immediate run 2
   real 2.19
   user 2.11
   sys 0.08

   end run 3
   real 2.13
   user 2.07
   sys 0.07

   immediate run 3
   real 2.14
   user 2.08
   sys 0.07
   ```

   Every end/immediate pair was byte-identical. Each output contained 5,835 lines and 405,631 bytes with SHA-256 `3ceb0ec056bf4e6e8ff4886cd4e71144f4e2267172fc3c2bffe5f0582b5c70ad`.
7. The separate `microsoft/TypeScript` `src/compiler` check used commit `cc5c6e2d32e2228fff83a66537bbe6042943054d`. The 40-rule aggregate output was byte-identical to the all-`end` baseline: 77 JSON lines, 1,148,510 bytes, SHA-256 `e7dc1eabe6996aad5876b2f4aae3c407935d068dbd8bed6271b0a811a6717a4d`. The individual `ts-namespace` probe did not pass equivalence: immediate traversal removed namespace containers and exposed nested declarations in `factory/utilities.ts`, `parser.ts`, and `checker.ts`. The same rule also changed `sst/opencode` output by exposing the nested `Service` interface in `packages/opencode/test/lib/llm-server.ts`, so both namespace mirrors remain `end`.

## How I checked it

```text
cargo test -p ast-grep-outline
58 passed; 0 failed

cargo test -p ast-grep test_outline
1 passed; 0 failed

cargo test --all-features --workspace
passed with zero failures

cargo fmt --all -- --check
passed

cargo clippy --all-targets --all-features --workspace -- -D clippy::all
passed

cargo clippy --all-targets --all-features --workspace --release --locked -- -D clippy::all
passed
```

Fixes #2825

## Summary by CodeRabbit

* **New Features**
  * Added configurable outline traversal boundaries with `stopBy: immediate` and `stopBy: end`.
  * Immediate matching now limits results to direct child items or members, while end matching continues through nested structures.
  * Updated TypeScript and TSX rules to use immediate matching for top-level declarations and imports/exports.

* **Bug Fixes**
  * Improved traversal pruning to avoid unnecessary processing of unrelated nested code.

* **Tests**
  * Added coverage for immediate and end matching across functions, classes, and mixed rule configurations.

### DIFF
--- a/crates/outline/src/combined_extractor.rs
+++ b/crates/outline/src/combined_extractor.rs
@@ -22,7 +22,9 @@ use ast_grep_core::{
 };
 use std::collections::HashMap;
 
-use crate::extractor::{ItemExtractor, MemberExtractor, OutlineRuleError, SerializableOutlineRule};
+use crate::extractor::{
+  ItemExtractor, MemberExtractor, OutlineRuleError, OutlineStopBy, SerializableOutlineRule,
+};
 use crate::model::{OutlineItem, OutlineMember};
 use crate::options::OutlineExtractorOptions;
 
@@ -35,6 +37,8 @@ pub struct CombinedExtractors<L: Language> {
   item_extractors: Vec<ItemExtractor<L>>,
   /// Dense node-kind index into `item_extractors`; shared across the whole file.
   item_kind_index: Vec<Vec<usize>>,
+  /// Whether every retained item extractor stops at direct children.
+  all_items_immediate: bool,
   /// Member extractors parsed once and referenced by parent-scoped groups below.
   member_extractors: Vec<MemberExtractor<L>>,
   /// Parent item extractor id to member extractors that may run inside it.
@@ -54,6 +58,24 @@ struct ScopedMemberExtractors<'a, L: Language> {
 struct MemberExtractorIndex {
   /// Sparse node-kind index into `member_extractors` for scoped member traversal.
   kind_mapping: HashMap<u16, Vec<usize>>,
+  /// Whether this parent scope contains an extractor that traverses to the end.
+  has_end: bool,
+}
+
+#[derive(Default)]
+struct TraversalVisitCounter {
+  #[cfg(test)]
+  count: usize,
+}
+
+impl TraversalVisitCounter {
+  #[inline]
+  fn record(&mut self) {
+    #[cfg(test)]
+    {
+      self.count += 1;
+    }
+  }
 }
 
 impl<L: Language> CombinedExtractors<L> {
@@ -105,10 +127,14 @@ impl<L: Language> CombinedExtractors<L> {
     options: OutlineExtractorOptions,
   ) -> Self {
     let item_kind_index = item_kind_index(&item_extractors);
+    let all_items_immediate = item_extractors
+      .iter()
+      .all(|extractor| extractor.common.stop_by == OutlineStopBy::Immediate);
     let member_index_by_parent = member_index_by_parent(&member_extractors);
     Self {
       item_extractors,
       item_kind_index,
+      all_items_immediate,
       member_extractors,
       member_index_by_parent,
       options,
@@ -125,7 +151,11 @@ impl<L: Language> CombinedExtractors<L> {
       })
   }
 
-  fn item_extractors_for_kind(&self, kind: u16) -> impl Iterator<Item = &ItemExtractor<L>> {
+  fn item_extractors_for_kind(
+    &self,
+    kind: u16,
+    allow_immediate: bool,
+  ) -> impl Iterator<Item = &ItemExtractor<L>> {
     self
       .item_kind_index
       .get(kind as usize)
@@ -133,6 +163,7 @@ impl<L: Language> CombinedExtractors<L> {
       .unwrap_or(&[])
       .iter()
       .map(|&idx| &self.item_extractors[idx])
+      .filter(move |extractor| allow_immediate || extractor.common.stop_by == OutlineStopBy::End)
   }
 
   pub fn extract<'a, 'tree>(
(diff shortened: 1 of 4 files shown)


HerringtonDarkholme:
This ticket does not accept external contribution, sorry

PR #2868, opened 2026-08-04
fix: match root metavariables against comments

### LINKED ISSUE #2867: Changed result matching comment in JS class member in ^0.45.0
### Discussed in https://github.com/ast-grep/ast-grep/discussions/2866

<div type='discussions-op-text'>

<sup>Originally posted by **donmccurdy** August  3, 2026</sup>
I'm trying to match any JSDoc comment on a JavaScript class member containing the JSDoc tag `@memberof`, and to remove that tag. In v0.44 I used the following successfully:

```yaml
id: no-jsdoc-memberof
message: >
  ES6 class members do not require @memberof tags.
severity: error
language: JavaScript
files:
  - packages/engine/Source/**/*.js
rule:
  kind: comment
  regex: "(?i)@memberof"
  pattern: $COMMENT
  inside:
    kind: class_declaration
    stopBy: end
transform:
  COMMENT_FIX: replace($COMMENT, replace="(?i)\n.*@memberof.*\n", by="\n")
fix: '$COMMENT_FIX'
```

As of v0.45, the rule no longer detects any issues on the following code block:

```javascript
class MyClass {
  /** @memberof MyClass.prototype */
  get myProp() { return 1; }
}
```

If I remove the `pattern: $COMMENT` line, the rule detects the issue again, but the transform no longer applies. I'm not 100% sure that I've understood some [relevant FAQ entries](https://astgrep.com/advanced/faq.html#kind-and-pattern-rules-are-not-working-together-why) so I'm not reporting this as a bug yet, but it seems unexpected to me to get a different result from v0.44 and v0.45. Any advice appreciated, thanks!

If this is intended, I can remove the transform. Just using ast-grep to detect the issue would be OK for my purposes, without necessary auto-fixing it.</div>

snowyukitty:
Fixes #2867.

Smart strictness skips extra comment nodes while matching structured patterns. That skip was also applied when the whole pattern was a metavariable, so a root `$COMMENT` could not match a comment candidate.

This keeps comment skipping for nested metavariables while allowing a Smart root metavariable to bind the candidate itself. Both normal matching and match-length calculation use the same root path.

Tests cover:

- named and dropped root metavariables on an extra TSX comment
- match-length consistency
- unchanged Relaxed behavior
- nested Smart metavariables continuing to skip comments

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Smart mode matching for root-level metavariables, ensuring they capture complete candidate nodes, including comments.
  * Preserved existing matching behavior for other root patterns.
  * Improved handling of dropped metavariables and strictness rules when comments or extra nodes are present.

### DIFF
--- a/crates/core/src/match_tree/match_node.rs
+++ b/crates/core/src/match_tree/match_node.rs
@@ -5,6 +5,25 @@ use crate::meta_var::MetaVariable;
 use crate::{Doc, Node};
 use std::iter::Peekable;
 
+pub(super) fn match_root_node_impl<'tree, D: Doc>(
+  goal: &PatternNode,
+  candidate: &Node<'tree, D>,
+  agg: &mut impl Aggregator<'tree, D>,
+  strictness: &MatchStrictness,
+) -> MatchOneNode {
+  // Smart matching ignores comments inside structured patterns, but a root
+  // metavariable represents the candidate itself and must still bind extras.
+  if let PatternNode::MetaVar { meta_var } = goal
+    && matches!(strictness, MatchStrictness::Smart)
+  {
+    return match agg.match_meta_var(meta_var, candidate) {
+      Some(()) => MatchOneNode::MatchedBoth,
+      None => MatchOneNode::NoMatch,
+    };
+  }
+  match_node_impl(goal, candidate, agg, strictness)
+}
+
 pub(super) fn match_node_impl<'tree, D: Doc>(
   goal: &PatternNode,
   candidate: &Node<'tree, D>,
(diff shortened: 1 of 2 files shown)


HerringtonDarkholme:
Thanks for the patch.

However, I need more time to think about if root pattern should be a special case.

matkoniecz:
note it was likely AI generated without adequate review if it is output of the same process as what ended as https://github.com/watchexec/watchexec/pull/1066#issuecomment-5193395417 was

HerringtonDarkholme (approved):

HerringtonDarkholme:
Thanks!

PR #2883, opened 2026-08-09
feat(outline): extract embedded language symbols

### LINKED ISSUE #2813: [feature] Add HTML support to `ast-grep outline`
# ⭐ Suggestion

It would be helpful if `ast-grep outline` could read HTML files. I'm not sure what the HTML structure would look like, but it could outline injected languages in `<script>` and `<style>` tags.

# 💻 Use Cases

We would find this especially useful for Vue files, which we can use with regular ast-grep with:

```yaml
languageGlobs:
  html:
    - "*.vue"
```

fyrsta7:
## Summary

- Extract outline items from embedded documents using the existing per-language outline rules.
- Merge host and injected items in source order while preserving their original file ranges.
- Add unit and CLI regression coverage for JavaScript and TypeScript in HTML and Vue files.

## Motivation

`ast-grep outline` previously only ran rules for the host document, so HTML files produced no symbols even when their `<script>` elements contained JavaScript or TypeScript declarations. This change reuses the existing language injection support so embedded declarations appear without inventing an outline structure for the parent HTML document.

Closes #2813.

## Testing

- `git diff --check`
- `cargo fmt --all -- --check`
- `cargo check`
- `cargo test`
- `cargo clippy --all-targets --all-features --workspace --release --locked -- -D clippy::all`
- `cargo test -p ast-grep --test run_test test_outline_javascript_in_vue_as_html`

## Summary by CodeRabbit

- **New Features**
  - Outline extraction now recognizes code embedded in supported languages within HTML-based files, including Vue components.
  - Combined outlines include items from both the host document and embedded scripts, sorted by their location in the source.
  - The compact JSON outline command now reports embedded functions, such as TypeScript or JavaScript functions, in its output.

### DIFF
--- a/crates/cli/src/outline/extract.rs
+++ b/crates/cli/src/outline/extract.rs
@@ -6,7 +6,8 @@ use std::sync::{Arc, mpsc};
 use std::thread;
 
 use anyhow::{Context, Result};
-use ast_grep_core::Language;
+use ast_grep_core::tree_sitter::StrDoc;
+use ast_grep_core::{AstGrep, Language};
 use ast_grep_language::LanguageExt;
 use ast_grep_outline::{
   DEFAULT_OUTLINE_RULES,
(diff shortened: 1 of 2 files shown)


fyrsta7:
Thank you for highlighting the uncovered path. I added a focused CLI regression test for HTML input via stdin and verified it with the full workspace coverage command. The previously uncovered extract_stdin line is now covered locally.

HerringtonDarkholme (approved):

PR #2895, opened 2026-08-18
feat(scan): add --severity threshold to filter findings by minimum level

### LINKED ISSUE #2720: [feature] Add --severity=<LEVEL> filter to `sg scan` to suppress lower-severity findings
## Summary

Add a `--severity=<LEVEL>` flag to `sg scan` that suppresses findings below the chosen threshold. Useful for CI logs where you want to see only errors (or only errors+warnings) without the noise of info/hint.

## Problem

Today `sg scan` reports every finding regardless of severity, and there is no way to suppress findings of a specific severity:

- `--error=RULE_ID` (and `--warning=...`, `--info=...`, `--hint=...`) **override** a rule's severity — they do not filter output. They also force a global change, not a per-run display filter.
- `--off=RULE_ID` **disables** a rule entirely; in a project with 30+ rules it is impractical to enumerate the rule IDs you want to silence for a single CI run.
- `--filter=REGEX` filters by **rule id**, not by severity. A project where warnings live under their own id namespace (`style/*`, `complexity/*`) cannot use it cleanly.
- `--max-results N` limits **count**, not severity. With 5,000 warnings, a CI log is still flooded.
- The only working pattern today is post-processing the output through `grep -E '^(error|warning)'` or similar — which is fragile and format-dependent.

In CI the typical need is: *"run all rules, but only print findings at error severity"* (or `error|warning`). This requires either disabling every non-error rule or piping through `grep`. Neither is good.

## Solution

Add a single flag:

```bash
sg scan --severity=error        # only error findings
sg scan --severity=warning      # warning + error findings
sg scan --severity=info         # info + warning + error findings
sg scan --severity=hint         # all findings (current default behavior)
```

Semantics: `--severity=LEVEL` is a **minimum-severity threshold**. Findings whose effective severity is below the threshold are not printed and do not contribute to the final tally or exit code.

The `Severity` enum already orders naturally (Hint < Info < Warning < Error) and is exposed as `ValueEnum` in the codebase, so the clap derivation is one line.

## Implementation sketch

In `crates/cli/src/utils/args.rs`, add a sibling filter to `OverwriteArgs` (or on `ScanArg` directly):

```rust
/// Only report findings at or above this severity level.
///
/// LEVEL ∈ hint | info | warning | error.
/// Findings below the threshold are not printed and do not affect
/// the final tally or exit code. Defaults to showing everything.
#[clap(long, value_name = "LEVEL")]
pub severity: Option<Severity>,
```

In `crates/cli/src/utils/rule_overwrite.rs` (or wherever `process_configs` filters the `RuleCollection`), drop any rule whose effective severity is below the parsed threshold. The `Severity` enum already implements `Ord` (or can be derived in 2 lines), and `RuleConfig<SgLang>` already has a `severity` field populated from the YAML.

Effect:
- `run_scan` only iterates over rules that survive the filter, so the worker pool never emits a suppressed finding.
- `error_count` (the atomic counter) only increments for findings at the chosen threshold and above, so the existing exit-code logic (`Err(EC::DiagnosticError(error_count))`) is automatically correct.
- `--json`, `--format=github`, `--format=sarif` and `--report-style=short|rich` all benefit from the same upstream filter — one change, four output surfaces.

## Use cases

1. **CI lint gate**: `sg scan --severity=error` in a job that should fail only on errors; warnings stay in the developer-facing `sg scan` (default) but don't break the build.
2. **PR review summary**: `sg scan --severity=warning --format=github` posts errors+warnings as a GitHub PR review comment and silences info/hint noise.
3. **Local "audit" mode**: `sg scan --severity=info` shows everything but skips `hint`, which often corresponds to stylistic nits the user already knows about.

## Distinct from existing flags

| Flag | Effect | Different because |
| --- | --- | --- |
| `--error=RULE_ID` etc. | Overrides severity | Mutates the rule, doesn't filter output |
| `--off=RULE_ID` | Disables rule | Removes the rule entirely; no per-severity granularity |
| `--filter=REGEX` | Selects rules by id regex | Filters by **id**, not by severity; can't say "all rules, only errors" |
| `--max-results N` | Stops after N findings | Limit on **count**, not severity |
| PR #2647 (in review) | Adds per-severity **count** to final tally | Orthogonal: shows the count of every level; doesn't suppress the per-finding output |

## Backward compatibility

- Default behavior (no flag) is unchanged.
- All existing flags compose with `--severity` (mutually orthogonal).
- No new public API, no rule YAML schema change, no breaking change to `Severity`.

## Testing

- Unit test in `crates/cli/src/utils/args.rs`: `severity` parses from each of the four level strings; `None` is the default.
- Integration test in `crates/cli/tests/`: a fixture repo with rules at each severity; assert that `--severity=warning` output contains `ErrorRule` and `WarnRule` lines but no `InfoRule` or `HintRule` lines; assert the final tally reflects only the kept levels; assert the exit code is 0 when only `HintRule` matches (without the flag it would be non-zero).

ChrisJr404:
## What

Adds `--severity=<hint|info|warning|error>` to `sg scan`: a minimum-severity **reporting threshold**. Findings from rules whose severity is below the threshold are not printed and do not contribute to the final tally or the exit code.

```bash
sg scan --severity=error     # only errors
sg scan --severity=warning   # warnings + errors
sg scan --severity=info      # info + warnings + errors
sg scan --severity=hint      # everything (current default)
```

Default behavior is unchanged (no flag = report everything).

## Why

Closes #2720. Today `sg scan` reports every finding regardless of severity, and there's no clean way to say "run all rules, but only surface errors (or errors + warnings)" in CI:

- `--error/--warning/--info/--hint` **override** a rule's severity, they don't filter output.
- `--off=RULE_ID` disables rules entirely and requires enumerating every id.
- `--filter=REGEX` filters by rule **id**, not severity.
- `--max-results N` limits **count**, not severity.

The only workaround is piping output through `grep`, which is fragile and format-dependent.

The semantics follow the convention you highlighted in the issue: ShellCheck's `--severity` and Biome's `--diagnostic-level` (minimum severity considered). The threshold is applied on the shared scan path, so text, JSON, SARIF and GitHub reporters all honor it consistently, and the existing exit-code logic stays correct because suppressed findings never increment the error count.

## How it's distinct from `--error/--warning/...`

| Flag | Effect |
| --- | --- |
| `--error=RULE_ID` etc. | Mutates a rule's severity |
| `--off=RULE_ID` | Removes a rule entirely |
| `--filter=REGEX` | Selects rules by id regex |
| `--severity=LEVEL` (this PR) | Filters output/tally by minimum severity, rules untouched |

## Implementation

- New `--severity` flag on `ScanArg`, parsed as a small `clap::ValueEnum` (`hint < info < warning < error`).
- A `passes_severity_filter` helper gates each `(rule, matches)` pair in both the path-based (`ScanWithConfig`) and stdin (`ScanStdin`) scan loops, before the `--max-results` accounting and the error tally.
- `Severity::Off` never produces findings, so it has no rank.

I chose the simplest, uniform interpretation (a rule below the threshold is not reported anywhere) rather than the "report-threshold but keep everything in JSON/SARIF" variant discussed in the thread. Happy to scope it to text/exit-code-only if you'd prefer that direction.

## Tests

- Unit test for `passes_severity_filter` covering every threshold/severity combination.
- Integration test `test_sg_scan_severity_threshold` asserting a `hint` rule is suppressed under `--severity warning` while a `warning` rule is still reported.
- Full `ast-grep` CLI suite passes; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

## Summary by CodeRabbit

* **New Features**
  * Added a `--severity` option for scan commands.
  * Set a minimum severity threshold to show only relevant findings.
  * Applied severity filtering consistently to filesystem and standard-input scans.
  * Filtered findings from displayed results, diagnostic counts, and error-based exit codes.
* **Tests**
  * Added coverage confirming default behavior and warning-level filtering.

### DIFF
--- a/crates/cli/src/scan.rs
+++ b/crates/cli/src/scan.rs
@@ -26,6 +26,41 @@ use crate::utils::{Items, PathWorker, StdInWorker, Worker};
 use std::collections::HashSet;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
+/// Minimum severity level accepted by the `--severity` reporting threshold.
+///
+/// Variants are ordered from least to most severe so that the derived `Ord`
+/// yields `Hint < Info < Warning < Error`.
+#[derive(clap::ValueEnum, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+enum SeverityFilter {
+  Hint,
+  Info,
+  Warning,
+  Error,
+}
+
+/// Map a rule's `Severity` to its reporting rank.
+///
+/// `Severity::Off` has no rank because disabled rules never produce findings.
+fn severity_rank(severity: &Severity) -> Option<SeverityFilter> {
+  match severity {
+    Severity::Hint => Some(SeverityFilter::Hint),
+    Severity::Info => Some(SeverityFilter::Info),
+    Severity::Warning => Some(SeverityFilter::Warning),
+    Severity::Error => Some(SeverityFilter::Error),
+    Severity::Off => None,
+  }
+}
+
+/// Whether a rule's findings should be reported under an optional minimum-severity threshold.
+///
+/// With no threshold every finding is reported, preserving the default behavior.
+fn passes_severity_filter(severity: &Severity, threshold: Option<SeverityFilter>) -> bool {
+  match threshold {
+    None => true,
+    Some(min) => matches!(severity_rank(severity), Some(rank) if rank >= min),
+  }
+}
+
 #[derive(Args)]
 pub struct ScanArg {
   /// Scan the codebase with the single rule located at the path RULE_FILE.
@@ -76,6 +111,16 @@ pub struct ScanArg {
   /// Useful for big codebase to fail scan/search fast.
   #[clap(long, conflicts_with = "interactive", value_name = "NUM")]
   max_results: Option<u16>,
+
+  /// Only report findings from rules at or above this severity LEVEL.
+  ///
+  /// LEVEL is one of hint, info, warning or error, in increasing severity.
+  /// Findings from rules whose severity is below the threshold are not printed
+  /// and do not contribute to the final tally or the exit code. This is a
+  /// minimum-severity threshold, mirroring ShellCheck's `--severity` and
+  /// Biome's `--diagnostic-level`. By default every finding is reported.
+  #[clap(long, value_name = "LEVEL")]
+  severity: Option<SeverityFilter>,
 }
 
 impl ScanArg {
@@ -267,6 +312,9 @@ impl PathWorker for ScanWithConfig {
         ret.push(processed);
       }
       for (rule, matches) in scanned.matches {
+        if !passes_severity_filter(&rule.severity, self.arg.severity) {
+          continue;
+        }
         // Atomically reserve slots for matches, truncating if needed
         let matches: Vec<_> = if let Some(counter) = &self.max_item_counter {
           let wanted = matches.len();
@@ -307,6 +355,7 @@ struct ScanStdin {
   // TODO: remove this
   error_count: AtomicUsize,
   max_diagnostics_shown: Option<usize>,
+  severity: Option<SeverityFilter>,
 }
 impl ScanStdin {
   fn try_new(arg: ScanArg) -> Result<Self> {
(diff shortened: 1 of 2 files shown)


HerringtonDarkholme:
Hey thanks for the contribution. However https://github.com/ast-grep/ast-grep/issues/2720#issuecomment-5245755459 needs more feedback for usage.

PR #2913, opened 2026-08-29
fix(outline): anchor fallback signatures to names

### LINKED ISSUE #2910: outline: decorated class/property declaration reports the leading decorator's line, not the declaration's own line
## Description

For a TypeScript/JavaScript class or property field with a leading decorator, `ast-grep outline` reports the entry's line and signature as the **decorator's own line**, not the actual `class`/field declaration line.

## Reproduction

```ts
// foo.ts
1: @Injectable()
2: export class Foo implements Bar {
3:   foo() {}
4: }
```

`ast-grep outline --json=pretty foo.ts`:
```json
{
  "role": "item",
  "symbolType": "class",
  "name": "Foo",
  "range": { "start": { "line": 0, "column": 0 }, "end": { "line": 3, "column": 1 } },
  "signature": "@Injectable()",
  "astKind": "export_statement"
}
```

Expected: `signature` should be `"export class Foo implements Bar {"` (or at minimum the class's own line), not the decorator's line.

Same issue affects decorated property fields (e.g. `@ApiProperty() email: string;` reports the decorator's line, not `email: string;`).

## Root cause (verified against source)

The matched node's span (`export_statement` for the decorated class) includes the leading decorator as part of the same node, and the outline extractor's signature fallback uses "the first non-empty line of the matched node" when no explicit `signature` template captures otherwise -- so the reported line/signature naturally lands on the decorator.

## Why a workaround via --outline-rules doesn't fix it

A custom outline rule targeting `class_declaration` directly (with an explicit `has: {field: name}` capture and `signature` template) does **not** override the bundled default extractor for the same span unless `--no-default-outline-rules` is also passed -- and that flag disables **all** bundled rules for the language, not just the one being overridden. Confirmed via `crates/cli/src/outline/extract.rs`'s `load_outline_rules()`: default and custom rules are loaded into a flat, unordered Vec with no id-based dedup/override/precedence logic -- both would independently emit their own entries for the same node, not replace each other.

## Suggested fix

Track the entry's signature line separately from the matched node's span start -- similar to the design in the (unrelated, coincidentally-named) [ast-outline](https://github.com/ast-outline/ast-outline) project, which hit an analogous bug and fixed it by adding a distinct `name_line` field independent of `start_line`. For `outline`, this could mean: when a bundled extractor's default fallback signature is used, skip past leading decorator/annotation-like leading children (if the language's grammar exposes them as identifiable node kinds) before taking "first non-empty line."

## Environment

- ast-grep 0.45.2 (current latest release as of 2026-08-26)
- Reproduced on macOS, TypeScript grammar

HerringtonDarkholme:
## Summary

- keep explicit outline signature templates authoritative
- anchor fallback signatures to NAME or an exact, source-backed name metavariable
- retain the first non-empty matched line when no source-backed name is available
- cover decorated TypeScript and JavaScript classes and fields

## Testing

- cargo test -p ast-grep-outline
- cargo clippy -p ast-grep-outline --tests -- -D warnings
- cargo fmt --all -- --check

Fixes #2910

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JavaScript and TypeScript outline extraction for declarations with decorators.
  - Decorated classes and fields now appear correctly in outlines, including exported classes and public properties.
  - Improved signature and anchor resolution for template-based matches, including exact-name and multi-match patterns.

- **Tests**
  - Added coverage for decorated JavaScript and TypeScript declarations.
  - Added validation for exact-name template matching and anchor resolution.

### DIFF
--- a/crates/outline/src/extractor.rs
+++ b/crates/outline/src/extractor.rs
@@ -12,7 +12,6 @@ use ast_grep_config::{
 use ast_grep_core::{
   Doc, Language, Node, NodeMatch,
   matcher::{Matcher, MatcherExt},
-  meta_var::MetaVariable,
   replacer::{Replacer, TemplateFix, TemplateFixError},
   source::Content,
 };
@@ -363,10 +362,7 @@ fn render_template<D: Doc>(template: &TemplateFix, node_match: &NodeMatch<D>) ->
   <D::Source as Content>::encode_bytes(&bytes).to_string()
 }
 
-fn default_signature<D: Doc>(
-  node_match: &NodeMatch<D>,
-  exact_name_var: Option<MetaVariable>,
-) -> String {
+fn default_signature<D: Doc>(node_match: &NodeMatch<D>, exact_name_var: Option<&str>) -> String {
   let node = node_match.get_node();
   if let Some(line) = signature_anchor_line(node_match, exact_name_var)
     && let Some(signature) = node_line(node, line)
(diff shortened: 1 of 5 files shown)


crates/outline/src/extractor.rs
```diff
@@ -362,9 +362,44 @@ fn render_template<D: Doc>(template: &TemplateFix, node_match: &NodeMatch<D>) ->
...
+      .get_match(name)
+      .map(|node| node.start_pos().line())
+      .or_else(|| {
+        env
+          .get_multiple_matches(name)
+          .first()
+          .map(|node| node.start_pos().line())
+      })
+  };
+  source_line("NAME").or_else(|| source_line(exact_name_var?))
```
HerringtonDarkholme: fake, no $NAME is default. remove the default NAME binding. if no exact_name_var, just fallback to first non-empty

PR #2939, opened 2026-09-15
feat: add riscv64 build support

threexc:
Make use of RISE's Native RISC-V Runners to build PyAV for riscv64. This requires enabling them for the repository as a GitHub Application. More info regarding enabling them is available here: https://riscv-runners.riseproject.dev/

I've tested these changes on my fork, where the only workflow to fail was coverage (due to a missing token): https://github.com/threexc/ast-grep/pull/1

This is being done on behalf of [RISE](https://riseproject.dev/), using the [RISC-V Wheels](https://stanfromireland.github.io/riscv-wheels/) dashboard to track upstream support for the Python ecosystem.

## Summary by CodeRabbit

* **New Features**
  * Added Linux wheel builds for RISC-V 64-bit systems.
  * Expanded package publishing coverage to include RISC-V alongside existing x86_64 and ARM64 targets.
  * RISC-V users can access prebuilt packages for supported Python environments, reducing the need to compile from source.
  * Existing x86_64 and ARM64 package builds remain supported.
  * Added target-specific build support for the appropriate Linux platform and Python architecture.


threexc:
I've added two extra commits to use the new `mirror:` option for `setup-python` as described by: https://github.com/actions/setup-python/issues/1288#issuecomment-5634506723

The alternative would be to use `setup-uv` (which is already used elsewhere in the workflows), since it supports riscv64 as well.

HerringtonDarkholme:
Hi, thanks for contributing.
Before supporting a new platform, I need to confirm there is real usage interest for ast-grep on that platform.

Can you share some context about using ast-grep-py on riscv64?

threexc:
> Hi, thanks for contributing. Before supporting a new platform, I need to confirm there is real usage interest for ast-grep on that platform.
> 
> Can you share some context about using ast-grep-py on riscv64?

Hi,

We don't have specific interest data for most packages. There's a chicken-and-egg problem in the Python ecosystem where adoption depends on support, but support depends on adoption. Projects which build with maturin (like ast-grep) make it very easy to get functional wheels out there for people to use, which is why I'm working on support across Rust-based packages specifically. 

I linked a dashboard showing status for the top 360 wheels on PyPI above, but we also have a separate RISE one covering almost 2,000: https://riseproject-dev.github.io/python-wheels-dashboard/. ast-grep-cli is listed there in light green, indicating we have it built and distributed in the RISE registry, but the secondary registry approach is really only a stopgap while upstream projects add support.

HerringtonDarkholme:
in general ast-grep follows Rust's tiered support. Tier1 platform will have their pypi/pyo3 bindings. For tier2 platforms, it requires at least user interest in using ast-grep on that platform.

Thanks for the contribution, the PR will be revisited if there are user interests in future.

Issue #2951, opened 2026-09-19
outline: Java enum constants are missing from java.yml


michal-michaluk:
## Summary

The bundled Java outline rules have no rule for `enum_constant`, so an enum's constants never reach the outline.

## Evidence

- Grammar: `tree-sitter-java` 0.23.5 `src/node-types.json` defines `enum_constant`.
- Schema: `schemas/java_rule.json` lists `enum_constant`.
- Rules: `crates/outline/src/default_rules/java.yml` has no rule matching it.

## Current behaviour (measured on `main`)

For:

```java
enum E { A, B; }
```

the outline is:

```
- Enum E
```

`A` and `B` are dropped.

## Proposed fix

A member rule scoped to the Java enum item, mirroring how other languages report enum members:

```yaml
id: java-enum-constant
language: Java
role: member
parentRuleIds: [java-enum]
symbolType: enumMember
rule:
  all:
    - kind: enum_constant
    - has:
        field: name
        pattern: $NAME
name: $NAME
isPublic: true
```

I have this prepared with a test; a PR follows.

PR #2954, opened 2026-09-19
feat(outline): add Java enum member rule

michal-michaluk:
Refs #2951.

## What this does

The bundled Java rules have no rule for `enum_constant`, so an enum's constants never reach the outline. This adds `java-enum-member`:

```yaml
id: java-enum-member
language: Java
role: member
parentRuleIds: [java-enum]
symbolType: enumMember
rule:
  all:
    - kind: enum_constant
    - has:
        field: name
        pattern: $NAME
name: $NAME
isPublic: true
```

Result:

```java
enum Colors { RED, GREEN; }
```

```
- Enum Colors
  - EnumMember RED
  - EnumMember GREEN
```

Before, only `- Enum Colors` was produced; `RED` and `GREEN` were dropped.

## Naming

`java-enum-member` follows the existing `csharp-enum-member`, `ts-enum-member`, `c-enum-member` and `cpp-enum-member`. (The issue draft wrote `java-enum-constant`; this matches the codebase convention instead.)

## Tests

- New `java_rules_parse_and_extract_enum_members`: an enum with constants and an empty enum.
- Updated `java_rules_parse_and_extract_jvm_surface`: it pinned the old behaviour, so the enum constant that now appears is expected.

`cargo test -p ast-grep-outline` green (9 tests), `cargo fmt --all -- --check` clean, `cargo clippy -p ast-grep-outline --all-targets -- -D warnings` clean.

## Summary by CodeRabbit

* **New Features**
  * Java enum constants are now recognized as public enum members.
  * Extracted Java outlines now include enum members, including members from named and empty enums.


michal-michaluk:
Re: `isPublic: true` on `java-enum-member`.

Kept intentionally, but you are right about the default, so here is the full picture for a maintainer call.

- The default is indeed `true`: `crates/outline/src/extractor.rs` compiles the member predicate as `compile_predicate(is_public, true)`, so `isPublic: true` is technically equivalent to omitting it.
- However, the bundled rules set it **explicitly** in **25** places (`grep -c '^isPublic: true' crates/outline/src/default_rules/*.yml`), and that includes **every analogous enum-member rule**:

  | rule | `isPublic` |
  | --- | --- |
  | `csharp-enum-member` | `true` |
  | `ts-enum-member` | `true` |
  | `c-enum-member` | `true` |
  | `cpp-enum-member` | `true` |

So dropping it here would make `java-enum-member` the only enum-member rule that relies on the default, which is the opposite of "consistent with the established rule convention".

I have kept `isPublic: true` for that reason. Happy to remove it if you prefer the other convention - your call.

